PR #3, opened 2017-09-12
Role management

endreman0:
- [x] `&giveme <rolename>[, <rolename>[, ...]]` gives you any number of roles in the giveme whitelist. Names are case-insensitive.
- [x] `&giveme add <rolename>` finds or creates a role named <rolename> and adds it to the giveme whitelist. Name is case-insensitive for existing role.
- [x] `&giveme create <rolename>` creates a role named <rolename> and adds it to the giveme whitelist.
- [x] `&giveme remove <rolename>[, <rolename>[, ...]]` removes any number of roles in the giveme whitelist from you. Names are case-insensitive.
- [x] `&giveme delete <rolename>` removes a role from the giveme whitelist and deletes it.
- [x] `&giveme list` lists all roles in the giveme whitelist.
- [x] `&give <member> <role>` gives <member> <role>. Members can be specified by username, username/discrim, nickname, ID, or mention, but are case-sensitive. Roles can be specified by name, ID or mention, but are case-sensitive.
- [x] `&take <member> <role>` removes <role> from <member>. Members can be specified by username, username/discrim, nickname, ID, or mention, but are case-sensitive. Roles can be specified by name, ID or mention, but are case-sensitive.

If you delete a giveable role through the role list rather than `&giveme delete`, its record will stay in the table but, it won't show up in `&giveme list` and will act like it's not there. Either 1) that's fine, 2) I add an event handler for when roles are deleted and delete the giveme record, or 3) I create `&giveme clean` that removes all IDs that don't exist at once. I'm impartial.

### DIFF
--- a/dozer/cogs/roles.py
+++ b/dozer/cogs/roles.py
@@ -1,5 +1,6 @@
 import discord, discord.utils
 from discord.ext.commands import bot_has_permissions, has_permissions, BadArgument
+from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
 from .. import db
 from ._utils import *
 
@@ -11,7 +12,6 @@ class Roles(Cog):
 		norm_names = [self.normalize(name) for name in roles.split(',')]
 		with db.Session() as session:
 			giveable_ids = [tup[0] for tup in session.query(GiveableRole.id).filter(GiveableRole.guild_id == ctx.guild.id, GiveableRole.norm_name.in_(norm_names)).all()]
-			
 			valid = set(role for role in ctx.guild.roles if role.id in giveable_ids)
 		
 		already_have = valid & set(ctx.author.roles)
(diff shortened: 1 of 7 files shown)


endreman0:
Shoot you're right, missed that. I'll get that fixed up in a second.

I used set because it's easier to do intersection and difference. Using plain lists, 17-18 would be
```py
already_have = [role for role in valid if role in ctx.author.roles]
given = [role for role in valid if role not in already_have]
```
which isn't as readable imo. Sets are also [wicked fast](https://wiki.python.org/moin/TimeComplexity#set) - those lines together run in `O(len(valid) + min(len(valid), len(ctx.author.roles)))` vs `O(len(valid) * len(ctx.author.roles) + len(valid) * len(already_have))`.

dozer/cogs/roles.py
```diff
@@ -0,0 +1,178 @@
+import discord, discord.utils
+from discord.ext.commands import bot_has_permissions, has_permissions, BadArgument
+from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
+from .. import db
+from ._utils import *
+
+class Roles(Cog):
```
Twinters007: where is Cog defined?

dozer/cogs/roles.py
```diff
@@ -0,0 +1,178 @@
...
+				raise BadArgument('that role already exists and is giveable!')
+			
+			role = await ctx.guild.create_role(name=name, reason='Giveable role created by {}'.format(ctx.author))
+			settings.giveable_roles.append(GiveableRole.from_role(role))
+		await ctx.send('Role "{0}" created! Use `{1}{2} {0}` to get it!'.format(role.name, ctx.prefix, ctx.command.parent))
+	
+	@giveme.command()
+	@bot_has_permissions(manage_roles=True)
+	async def remove(self, ctx, *, roles):
+		"""Removes multiple giveable roles from you. Names must be separated by commas."""
```
Twinters007: you should add that it can contain spaces in the separator as well

dozer/cogs/roles.py
endreman0: Cog is defined in `cogs/_utils.py`, and imported by `from ._utils import *`.

dozer/cogs/roles.py
endreman0: I asked a sample of potential users for an example of a comma-separated list, and their responses include spaces. I believe it's common knowledge. Anybody else want to chime in?

Twinters007: Explicit is better than implicit

dozer/cogs/roles.py
Twinters007: explicit is better than implicit

PR #8, opened 2017-09-17
Ban, Unban, Kick

tweirtx:
Now that I've made a few improvements, here we go!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 __pycache__
 config.json
+dozer.db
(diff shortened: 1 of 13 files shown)


mcao:
Edited a bit, fine for right now, though I would implement checks to see if the bot has BAN_MEMBERS/KICK_MEMBERS and if it has the needed perms to kick that member, as well as check if the person executing the command has the perms to ban/kick the bannee/kickee.

PR #9, opened 2017-09-18
Perms are important

tweirtx:
It was pointed out to me that mod functions should probably check if the initiator actually is allowed to use them. I also added back a feature that was removed because I wasn't clear on the purpose.

### DIFF
--- a/dozer/cogs/moderation.py
+++ b/dozer/cogs/moderation.py
@@ -27,6 +27,8 @@ class Moderation(Cog):
     @command()
     async def kick(self, ctx, user_mentions: discord.User):
         "Kicks the user mentioned."
+        haspermissions = ctx.channel.permissions_for(ctx.message.author)
+        print(haspermissions)
         usertokick = user_mentions
         usertokickstr = str(usertokick)
         await ctx.send("Kicking " + usertokickstr)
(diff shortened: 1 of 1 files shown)


tweirtx:
I could not get the perms to work correctly. I tried everything I could think of on my personal server. I don't know why manage_guild is the only one that seems to work.

tweirtx:
@mcao Thanks to Ian, the issue has now been fixed. Should be good to go now.

tweirtx:
For some reason the permissions weren't working properly for me but now they should be good to go. Not sure why it wasn't working for me but it's working under Ian's instance.

mcao:
Approved for now, though I would encourage you to work on putting a reason into Discord's built in mod-logs as well as logging it the way we do in #mod-logs in the FRC server. Also, througher permissions checks.

tweirtx:
@bkeeneykid you have questions, I have answers.
1. Valid point, I'll be sure to change that in the next revision.
2. Discord.py or Ian's framework (not sure which) actually handles that automatically, so no need to write my own code for that.
3. Not sure what you mean by that, please provide more detail.
4. I didn't do that, that was Ian that did it.

@mcao To answer your feedback,
1. It's on my list.
2. Not sure it can get more thorough than that, either the command runner can ban or kick or they can't.

endreman0:
tweirtx is correct, discordpy handles argument parsing automatically. The argument will never be `None` - no argument provided will raise a `MissingRequiredArgument`, and invalid argument (wrong format or nonexistent/unknown user) will raise a `BadArgument` before the command is called.

That "extra space" is GitHub's rendering. Apparently tabs are rendered as 3 spaces. I converted indentation to tabs to match the rest of the codebase.

PR #12, opened 2017-09-20
Rewritten help command

endreman0:
- [x] Add `&help`, a command to show information about the bot, a specific command, or a cog
- [x] Alias `&about` to `&help`, as `&help` now contains all of the information from `&about` and more
- [x] Add custom command and group classes with example usage
- [x] Add help and example usage for every command
- [x] Add help information for cogs

### DIFF
--- a/dozer/cogs/general.py
+++ b/dozer/cogs/general.py
@@ -15,7 +15,11 @@ class General(Cog):
 		delay = response.created_at - ctx.message.created_at
 		await response.edit(content=response.content + '\nTook %d ms to respond.' % (delay.seconds * 1000 + delay.microseconds // 1000))
 	
-	@command(name='help')
+	ping.example_usage = """
+	`{prefix}{name}` - Calculate and display the bot's response time
+	"""
+	
+	@command(name='help', aliases=['about'])
 	async def base_help(self, ctx, *target):
 		"""Show this message."""
 		if not target: # No commands - general help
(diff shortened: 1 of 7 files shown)


dozer/cogs/info.py
```diff
@@ -6,10 +6,21 @@
...
+	async def member(self, ctx, member: discord.Member=None):
+		"""
+		Retrieve information about a member of the guild.
+		If no arguments are passed, information about the author is used.
+		**This command works without mentions.** Remove the '@' before your mention so you don't ping the person unnecessarily.
+		You can pick a member by:
+		- Username (`cooldude`)
+		- Username and discriminator (`cooldude#1234`)
+		- ID (`326749693969301506`)
+		- Nickname - must be exact and is case-sensitive (`"Mr. Cool Dude III | Team 1234"`)
```
bkeeneykid: I might suggest using the `fuzzywuzzy` library like we did earlier for the feedback module, and finding the one with the greatest similarity. At the very least, keep it case insensitive.

bkeeneykid: Nevermind, I just realized this is discord's built in user finder. Ignore me.

dozer/cogs/general.py
```diff
@@ -11,6 +14,104 @@ class General(Cog):
...
+	
+	@command(name='help', aliases=['about'])
+	@bot_has_permissions(add_reactions=True)
+	async def base_help(self, ctx, *target):
+		"""Show this message."""
+		if not target: # No commands - general help
+			await self._help_all(ctx)
+		elif len(target) == 1: # Cog or command
+			target_name = target[0]
+			if target_name in ctx.bot.cogs:
```
bkeeneykid: I might suggest searching for commands before cogs, as if you have a guilds cogs and a `&guilds`, the cogs documentation will be less than useful. Better yet, dynamically display both at the same time if they're named the same thing.

PR #21, opened 2017-10-07
UX and DX quality-of-life changes

endreman0:
UX and DX being User and Development eXperience, respectively.
- [x] Prevent commands with no help message from breaking the help command
 - If usage is provided, suggest using help for that command to see usage
 - If not, "No information provided."
- [x] Silence CommandNotFound errors
- [x] Ignore messages by bots, and multiple messages from the same user sent within one second
- [x] Apply cooldowns to help, guild, member, and giveme list, as they are especially spam susceptible
- [x] Prevent privilege escalation through give and take commands: attempting to give or take a role higher than the caller's top role will raise an error
- [x] Fix all subcommands being groups, even if decorated with `@command` instead of `@group`
- [x] Allow database sessions to be used as asynchronous context managers. Many commands inserting or grabbing many records from the database will take long enough to warrant `ctx.typing()`, and allowing use as an asynchronous context manager streamlines that code.
```py
# current code
async with ctx.typing():
    with db.Session() as session:
        # do expensive database work
# send results

# new code (this PR)
async with ctx.typing(), db.Session() as session:
    # do expensive database work
# send results
```

### DIFF
--- a/dozer/bot.py
+++ b/dozer/bot.py
@@ -32,6 +32,8 @@ class Dozer(commands.Bot):
 		elif isinstance(err, commands.BotMissingPermissions):
 			permission_names = [name.replace('guild', 'server').replace('_', ' ').title() for name in err.missing_perms]
 			await ctx.send('{}, I need {} permissions to run this command!'.format(ctx.author.mention, utils.pretty_concat(permission_names)))
+		elif isinstance(err, commands.CommandOnCooldown):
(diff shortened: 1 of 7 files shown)


endreman0:
@bkeeneykid With the `has_permissions` and `bot_has_permissions` decorators, the bot will already show what permissions it's missing [here](https://github.com/FRCDiscord/Dozer/blob/79237258cc023bb292ba35bbec51ba9b91c91db5/dozer/bot.py#L29-L34). If the command raises a Forbidden, there's no information in the error to figure out what failed and why, which is why those decorators exist.

PR #26, opened 2017-10-16
Example and Fix for Timezone command

770352:


770352:
Sorry, Forgot to push code.

770352:
I did use .format

endreman0:
There's string concatenation on line 97. While this isn't strictly your PR, for fixing general issues with the timezone command we might as well lump it in.

770352:
There you go.

PR #28, opened 2017-10-19
Implement a prune command

guineawheek:


tweirtx:
[prune command for bulk deletes in a channel](https://trello.com/c/baFpezst/41-prune-command-for-bulk-deletes-in-a-channel)

PR #29, opened 2017-10-21
Added a New Member command

LeoPiGuy:
Added `{prefix}nmconfig` command to the moderation cog with the appropriate help information.


LeoPiGuy:
Spelling should be fixed @tweirtx

LeoPiGuy:
@bkeeneykid Yea sure I can add that to my cleanup PR if you'd like.

endreman0:
@bkeeneykid not even one cog; excalibot does it in [one command](https://github.com/endreman0/Excalibot/blob/db-rewrite/excalibot/cogs/config.py). (The multiple commands are for user settings vs guild settings, and a developer-only command to set any column in any record.) It's a bit dirty now - it only works for bool, float, int, None, and str right now, and there's no way to prevent putting a channel ID in a role column or vice versa - but

LeoPiGuy:
@guineawheek Sounds good. I'm thinking maybe finish this PR first then add that functionality. At the time, having a basic new member system for the main FRC server is in my mind more important then those formalities. Ill see if @endreman0 agrees. Either way I will work with Travis on this.

endreman0:
@guineawheek while 1 & 2 are important for the FRC guild, there are many smaller guilds that have different or no systems for new members. A configurable message to DM new members, and/or a configurable pattern for new members' nicknames (i.e. `{username} | SET TEAM#`), would be useful, but I agree with @LeoPiGuy that getting the basics first and the goodies later is more important than dropping everything in at once.

PR #33, opened 2017-10-23
Made &nick work better with names over 32 characters

guineawheek:
Now, instead of outputting an ugly api error, it simply truncates the nickname and warns the user of the truncation!


tweirtx:
[/nick command for mobile users to set their nickname](https://trello.com/c/wupFheRX/9-nick-command-for-mobile-users-to-set-their-nickname)

PR #34, opened 2017-10-23
Made the invite command

770352:


770352:
Nah Fam, Thats too easy.

770352:
Should be all fixed.

770352:
Or, I can just hard code it and it will be fine. Quit over complicating things.

endreman0:
Overcomplicating things is leaving the bot's permissions as a hardcoded number, and generating more work down the line. Whenever we add a command, in this case, we have to find the numeric value of every permission the command needs and make sure it is included in that number, which is tedious and error-prone. Hard-coding will not be fine.

tweirtx:
I have been told by Harold that he has abandoned this command. Closing PR.

PR #36, opened 2017-10-27
Timeout command

endreman0:
- `timeout` command to disable sending message and adding reactions
  - If no member role is set, every role under the caller's top role is muted and the caller is warned
  - If a role is set, that role is muted
  - The bot and the caller can send messages, to give everyone a stern talking-to while they're listening
- `memberconfig` command to set a members role
  - This role is muted during a timeout
  - Cannot be the caller's top role or any role higher than it
  - This role should be the most inclusive role that gives send message and add reaction permissions
  - To set the default role as the members role, use any of `@everyone`, `@.everyone`, `@/everyone`, `@ everyone`, or `@​everyone` (with a zero-width space)

### DIFF
--- a/dozer/cogs/moderation.py
+++ b/dozer/cogs/moderation.py
@@ -142,10 +142,13 @@ class Moderation(Cog):
 		Set the member role for the guild.
 		The member role is the role used for the timeout command. It should be a role that all members of the server have.
 		"""
+		if member_role >= ctx.author.top_role:
+			raise BadArgument('member role cannot be higher than your top role!')
+		
 		with db.Session() as session:
 			settings = session.query(MemberRole).filter_by(id=ctx.guild.id).one_or_none()
 			if settings is None:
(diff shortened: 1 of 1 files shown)


endreman0:
Oops, you're right, I removed that decorator accidentally. Good catch! Fixed.

endreman0:
I see that that's included in PEP8. For this PR I would prefer to leave it, as there are other instances of it ([1], [2], [3], and every cog except `moderation.py` and `tba.py`) and internal consistency overrides all. I think a separate PR to clean up styling, and cover all modules at once, would be the better solution.

[1]: https://github.com/FRCDiscord/Dozer/blob/master/dozer/utils.py#L1
[2]: https://github.com/FRCDiscord/Dozer/blob/master/dozer/bot.py#L1
[3]: https://github.com/FRCDiscord/Dozer/blob/master/dozer/__main__.py#L1

PR #42, opened 2017-11-02
Implement the voicebind command and automatic role giving/taking

guineawheek:
yay got around to the db stuff (finally).

Seems to be fully working atm.


mcao:
Have you accounted for moving from one voice channel to another?

guineawheek:
Fixed the switching between channels thing, oops.

Also updated for Devin's suggestion as well

guineawheek:
That's why there's a check to make sure that the channels between the
before and after states are actually different as well

tweirtx:
[Voice text channel handling](https://trello.com/c/Vz8N57Bc/36-voice-text-channel-handling)

guineawheek:
Alrighty, I added in some docs for `voicebindlist`

PR #45, opened 2017-11-10
Add stopgap TOA functionality

guineawheek:
This implements a basic toa command that has feature parity with where Robot-v5 was previously. I wasn't intending to submit this PR as it was mostly intended for the FTC server's Dozer, but Travis asked me to so here it is.

Please note that down the line `_toa.py` should be ostensibly be replaced with Ian's `toapy` down the line, as it's an awful hack - it simply stuffs the raw JSON data it gets into the attribute dictionary of what's essentially a data class, without doing any coercion to Pythonic types beyond what `json.loads` would give you.

### DIFF
--- a/dozer/__main__.py
+++ b/dozer/__main__.py
@@ -1,7 +1,7 @@
 import json, os.path, sys
 from . import db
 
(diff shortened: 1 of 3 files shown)


guineawheek:
That might help. Fixed.

PR #46, opened 2017-11-17
Added a command to remove a role from the giveme list without deleting the role from the server

tweirtx:


tweirtx:
Whoops haha thanks...will fix soon

PR #50, opened 2017-11-26
Speedtest command and addition of web socket latency to ping command

770352:
Uses speedtest-cli to run a speed test from [http://www.speedtest.net](url). The use for this would be for quick and remote diagnostics if the bot has high latency. The command will immediately return the following in an embed,
```
Please wait
Running speedtest, This may take a couple mintues.
```
Once the speed test is compelete, The bot will remove the Please Wait message and return the Results in the following format,
```
Speed test complete
Ping: [number] ms
Download: [number] Mbit/s
Upload: [number] Mbit/s```


770352:
Changes complete, Invite command removed.

770352:
1. I got nothing for that.
2. Not the way I have it setup. (Check Dozer [H2020])
3. I kinda had a brainfart there.

770352:
3 is fixed, 2 was never broken.

770352:
1 is met, 2 is met and 3 is met.

770352:
I don't get you need it done one exact way, The way I'm doing it works and does the exact same thing that the way you think it should be done does it, I just achieve it in a different way.

770352:
I will remove the speed test command for this PR, But I will have to insert a entry into .gitignore.

770352:
@guineawheek That is a file that everyone can use for any custom code on Dozer that is either not ready to be pushed or for people who need it for Team Servers and whatever else custom.

guineawheek:
I have one of those. It's called `hacks.py`. You never see it in commits _because I don't stage it, like, ever._ Your PR will not be merged until that is fixed.

770352:
Give me a few and I will work on it.

770352:
All should be good now.

770352:
It was susposed to have a speedtest command and now Ian just wants a Webhook latency thing added to the ping command.

TomOrth:
But still not meaningful enough to be commited imo. Considering it has ~29 commits

endreman0:
To clarify on things being done "one exact way": there are a variety of ways of determining the latency of a connection without:
- long-running, blocking code (running &speedtest effectively disables the bot until the test is complete)
- consuming additional bandwidth to run the test
- opening a potential DoS vector by executing the command several times concurrently (assuming the blocking issue is fixed)

I would accept any solution which passes some reasonable checks such as those. The speedtest command as written was not one of them.

PR #52, opened 2017-11-29
Removing FTCRoot before FTCRoot closes

770352:


guineawheek:
As a followup, I also discovered that this PR ADDS THE BINARY INSTALLERS FOR PYTHON AND GIT THEN REMOVES THEM!! WHICH WOULD MAKE DOWNLOADING DOZER FAR SLOWER!! AAAAAAAGH

From now on, I'm blocking every PR from this branch until the OP burns it to the ground and starts over on a new one based on the current master.

WardBenjamin:
Agreed. Every branch should be forked clean from current master to avoid extraneous changes and prevent merge conflicts as much as possible. Branches REALLY SHOULD NOT contain commits from other PRs.

PR #53, opened 2017-11-29
Make the TOA command fallback to a static dump of FIRST's DB when TOA is missing data. 

guineawheek:
See title. This fixes the currently borked rookie year field, adds a website field for most teams, and can display dead teams that aren't currently in TOA. 

Additionally, this replaces the FTCRoot field with a TOA link.


guineawheek:
@WardBenjamin You _could_, but it would likely come at the cost of increased latency. Perhaps a better approach is to use an actual database. Then again, most of the TOA code in place atm is a nasty hack that's only in place because TOA and toapy (as of yet) are incomplete. 

Given that Dozer's memory usage (according to htop), is already around 100MB anyway, I'm not too concerned.......

PR #55, opened 2017-12-01
Made self-deafens unable to @everyone in modlogs

tweirtx:


tweirtx:
[Make self-deafens not @everyone](https://trello.com/c/STeUbqHC/57-make-self-deafens-not-everyone)

tweirtx:
After discussion, we're just gonna make modlogging one huge function that handles everything automatically

PR #63, opened 2017-12-19
Bug fixes

tweirtx:
Since it was my stupidity that introduced some bugs, I felt responsible for fixing them. Here are those fixes. This will fix the incorrect user being displayed in &mute and &deafen as well as selfdeafens not undoing themselves.


tweirtx:
[Moderation display error](https://trello.com/c/u3zREtKu/60-moderation-display-error)

tweirtx:
[Self deafens don't undeafen](https://trello.com/c/eBbaZreP/61-self-deafens-dont-undeafen)

PR #65, opened 2017-12-27
Team number database

tweirtx:
This is a way to see who's associated with what team(s) as well as what team(s) a particular person is associated with

### DIFF
--- a/dozer/cogs/teams.py
+++ b/dozer/cogs/teams.py
@@ -3,40 +3,26 @@ from .. import db
 from ._utils import *
 import discord
 
-# noinspection PyUnboundLocalVariable
-
 
 class Teams(Cog):
 	@command()
 	async def setteam(self, ctx, team_type, team_number):
 		team_type = team_type.casefold()
 		with db.Session() as session:
-			if team_type == "frc":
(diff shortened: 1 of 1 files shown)


tweirtx:
I think they should as more commands relating to the teams DB may follow. Not confirmed yet but there is a possibility

tweirtx:
[Team number DB](https://trello.com/c/kkRnuxmo/51-team-number-db)

dozer/cogs/teams.py
```diff
@@ -0,0 +1,85 @@
...
+	@command()
+	async def setteam(self, ctx, team_type, team_number):
+		"""Sets an association with your team in the database."""
+		team_type = team_type.casefold()
+		with db.Session() as session:
+			dbtransaction = TeamNumbers(user_id=ctx.author.id, team_number=int(team_number), team_type=team_type)
+			session.add(dbtransaction)
+		await ctx.send("Team number set!")
+	setteam.example_usage = """
+	`{prefix}setteam type team_number` - Creates an association in the database with a specified team
```
bkeeneykid: For the team type, do we have a standard system for this? I understand most people say "FRC" and "FTC" but what about VRC? Won't some people say vex or something else? We should have the accepted types in the example usage and check if it's in there.

PR #64, opened 2017-12-24
A rewrite of the namegame bot, integrated into Dozer

guineawheek:
This is a mostly complete reimplementation of Devin's original namebot. 

Featuring:
 * Separate timer event loop with game state locks to try to avoid game state corruption and crashes
 * FRC and FTC support
 * More embeds that count down 

Not featuring:
 * Any sort of database stuff yet (leaderboard, config options)

PR #66, opened 2017-12-27
New and improved namegame pull request

guineawheek:
Featuring:
* fixes for really stupid bugs the previous PR had like not actually checking for correct starting digit
* refactoring to remove the giant loop, and reaction handling is actually event based rather than polled
* configurability and leaderboards so you can’t disrupt #general with a game

Not featuring:
* things that will probably end up in a new PR
* proper help for the ng config subcommands, please comment on how i should tweak it


guineawheek:
Reopening #64

PR #67, opened 2018-01-01
Upgrade to TBApi v1.3.1b2 and TBA API v3

endreman0:
Caching is disabled temporarily, as it is not yet implemented in TBApi.


tweirtx:
Since this falls under emergency patch protocol, only 1 review is required and I am merging this

PR #79, opened 2018-01-12
Automatic team number appending to nickname on join

tweirtx:
This will append a team number to a user's nickname when they join if they have exactly 1 team association in the database.

### DIFF
--- a/dozer/cogs/teams.py
+++ b/dozer/cogs/teams.py
@@ -79,7 +79,9 @@ class Teams(Cog):
 			with db.Session() as session:
 				query = session.query(TeamNumbers).filter_by(user_id=member.id).all()
 				if len(query) == 1:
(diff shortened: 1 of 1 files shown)


tweirtx:
[Automatic nickname setting on entry verification w/ team num](https://trello.com/c/oeYFkknv/13-automatic-nickname-setting-on-entry-verification-w-team-num)

WardBenjamin:
Just as a note, `query.first()` can be MASSIVELY faster than `query.all()` because the SQL statement immediately exits when it finds a match rather than searching the database. (It's an identical speed if there's no row that matches, but otherwise it'll improve speeds in almost all cases.)

tweirtx:
I could have sworn I changed that, let me look into it

tweirtx:
As well as the nickname thing

PR #80, opened 2018-01-16
Allow Dozer to delete posts with links if a user lacks a particular role

guineawheek:
this is meant as a spam control tool and is semi-urgent for one of dozer’s servers


endreman0:
My mobile app doesn't allow reviewing PRs, but this looks good. You are good to merge.

PR #82, opened 2018-01-20
Handled people associating with the same team twice throwing an error

tweirtx:


tweirtx:
[Fix &setteam throwing an SQL error when someone tries to associate themselves with the same team twice](https://trello.com/c/MkzX9ap3/66-fix-setteam-throwing-an-sql-error-when-someone-tries-to-associate-themselves-with-the-same-team-twice)

PR #83, opened 2018-01-21
Update onboarding to require nickname includes team number

mehnadnerd:
The rules should also be updated to indicate this procedure

### DIFF
--- a/dozer/cogs/moderation.py
+++ b/dozer/cogs/moderation.py
@@ -147,12 +147,16 @@ class Moderation(Cog):
 				string = config.message
 				content = message.content.casefold()
 				if string not in content: return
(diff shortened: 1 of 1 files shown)


mehnadnerd:
Now urges volunteers to set team number to 0000, and doesn't require people to put 0000 in their names. I'm not sure what the issue is with the Vex people. If the number in the database is part of their name, it will allow it, even if it has trailing letters.

endreman0:
Guinea is referring to the fact that VEX teams are some number of digits followed by a letter. That's an issue with the setteam command, not this PR, but requiring a team association to join any server will make it a problem that VEX guilds can't ignore.

tweirtx:
Closing due to inactivity.

PR #88, opened 2018-02-12
TBA embeds now support avatars

tweirtx:


tweirtx:
[FRC Avatars in TBA embeds](https://trello.com/c/dqPBw49Q/77-frc-avatars-in-tba-embeds)

PR #91, opened 2018-03-21
fix things in moderation.py and namegame.py

guineawheek:
no more namegame timer deaths....


tweirtx:
[namegame voting and timers stop after skipping](https://trello.com/c/yDLVVaCu/74-namegame-voting-and-timers-stop-after-skipping)

Issue #77, opened 2018-01-06
Follow PEP8


Redrield:
I forked dozer to contribute it, opened PyCharm and this is the first thing that I see when I open a file
https://imgur.com/aNDQYOP

I feel like just a general refactor to follow PEP8 would help in terms of consistency and such

PR #96, opened 2018-04-14
Reformat everything to be much more pep8 compliant

guineawheek:
Basically run through pycharm's reformatter and some glaring naming issues fixed (mostly in the moderation cog)

There's def more stuff not in pep8 compliance (mostly long lines), and much room for code cleanup, but since this is the main PR which switches Dozer from tabs to spaces, the faster this gets done and over with, the less merge conflicts happen....

oh btw this addresses #77

### DIFF
--- a/dozer/__main__.py
+++ b/dozer/__main__.py
@@ -1,41 +1,44 @@
-import json, os.path, sys
+import json
+import os.path
+import sys
+
 from . import db
 
 config = {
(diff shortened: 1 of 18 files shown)


guineawheek:
Admittedly it's one of the things I dislike about PEP8 - long lines aren't always more readable as prescribed by it

PR #101, opened 2018-04-22
fixed &user to work with newer versions of discord.py rewrite

guineawheek:
a little hacky (because it strips off `ActivityType.playing` to `playing` etc) but it works

breaks compat with older discord.py tho


guineawheek:
fixed listening bug, please re-review

PR #102, opened 2018-04-28
Giveme roles now no longer list when deleted from server

tweirtx:
There have been multiple instances where  I've had to log into the DB to remove roles that have been deleted from the guild but not the giveme list, and this PR aims to put an end to that.

### DIFF
--- a/dozer/cogs/roles.py
+++ b/dozer/cogs/roles.py
@@ -15,8 +15,8 @@ class Roles(Cog):
         for command in self.giveme.walk_commands():
             @command.before_invoke
             async def givemeautopurge(self, ctx):
-                if await self.ctxpurge(ctx, False):
-                    await ctx.send("That role does not exist any more, removing from list")
+                if await self.ctxpurge(ctx):
+                    await ctx.send("Purged missing roles")
 
     async def on_member_join(self, member):
         me = member.guild.me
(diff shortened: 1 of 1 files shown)


dozer/cogs/roles.py
```diff
@@ -68,6 +76,25 @@ class Roles(Cog):
...
+        with db.Session() as session:
+            roles = session.query(GiveableRole).filter_by(guild_id=ctx.guild.id)
+            guildroles = []
+            for i in ctx.guild.roles:
+                guildroles.append(i.id)
+            for role in roles:
+                if role.id not in guildroles:
+                    await self.givemepurge(role)
+                    counter += 1
+        if returncounter:
```
endreman0: Because everything is truthy/falsy, you can return the counter always, and it will work for if-checks.

dozer/cogs/roles.py
```diff
@@ -68,6 +76,25 @@ class Roles(Cog):
...
+    async def ctxpurge(self, ctx, returncounter):
+        counter = 0
+        with db.Session() as session:
+            roles = session.query(GiveableRole).filter_by(guild_id=ctx.guild.id)
+            guildroles = []
+            for i in ctx.guild.roles:
+                guildroles.append(i.id)
+            for role in roles:
+                if role.id not in guildroles:
+                    await self.givemepurge(role)
```
endreman0: This still creates a new transaction for every role deleted, when deleting all missing roles could be one query. Given a list of all the role IDs in a guild, you can find records for that guild whose role ID is [not in](http://docs.sqlalchemy.org/en/latest/core/sqlelement.html#sqlalchemy.sql.expression.ColumnElement.notin_) that collection, then [delete](http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.delete) them all at once.

dozer/cogs/roles.py
```diff
@@ -5,11 +5,19 @@
...
 
 class Roles(Cog):
     """Commands for role management."""
 
+    def __init__(self, aa):
+        for command in self.giveme.walk_commands():
+            @command.before_invoke
+            async def givemeautopurge(self, ctx):
+                if await self.ctxpurge(ctx, False):
+                    await ctx.send("That role does not exist any more, removing from list")
```
endreman0: "That role" wouldn't make much sense for `&giveme list` or other commands that don't specify a role; and if, for example, a user ran `&giveme a, b` and the bot deleted a missing record that the user didn't request, seeing both "That role doesn't exist" and "Gave you two roles" indicates it both did and didn't complete successfully. "Purged {n} missing role(s)" would make more sense to send.

PR #105, opened 2018-05-10
Fix for NameGame Team Max Error

bkeeneykid:
A quick and dirty patch to split up the picked teams into more than one embed value if necessary.


tweirtx:
[Namegames can go for long enough that team listing breaks because the embed is too big](https://trello.com/c/cWpVDwEc/90-namegames-can-go-for-long-enough-that-team-listing-breaks-because-the-embed-is-too-big)

PR #108, opened 2018-05-13
fixed some issues with embeds and check_links in moderation.py

guineawheek:


guineawheek:
@endreman0 I have no idea but somehow it happens in practice:

```
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/discord/client.py", line 224, in _run_event
    yield from coro(*args, **kwargs)
  File "/var/lib/dozer-ftc/run/dozer/cogs/moderation.py", line 193, in on_message_edit
    await self.check_links(after)
  File "/var/lib/dozer-ftc/run/dozer/cogs/moderation.py", line 109, in check_links
    if role not in msg.author.roles and re.search("https?://", msg.content):
AttributeError: 'User' object has no attribute 'roles'
```

My guess is webhooks.

PR #110, opened 2018-05-15
Invite command

endreman0:
Keeping track of what permissions the bot requires involved replacing the default `bot_has_permissions` decorator with one that sets an attribute on the command with the required permissions. This also required putting `CommandMixin` at the front of `Command` and `Group`'s base classes, since discordpy's `Command` and `Group` don't call the superclasses' `__init__` methods.

### DIFF
--- a/dozer/cogs/_utils.py
+++ b/dozer/cogs/_utils.py
@@ -1,14 +1,27 @@
 import asyncio
 import inspect
+import discord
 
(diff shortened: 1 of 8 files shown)


tweirtx:
[Invite command](https://trello.com/c/HyvmSfd5/52-invite-command)

PR #111, opened 2018-05-17
Fix NameGame to check the full name of the team

bkeeneykid:


tweirtx:
[you can get away with putting in "o" and having that be 100% correct on NameGame](https://trello.com/c/2GjOfQC2/75-you-can-get-away-with-putting-in-o-and-having-that-be-100-correct-on-namegame)

PR #114, opened 2018-07-03
Add word filter

bkeeneykid:
This is a mostly complete word filter. It's fully configurable about the filters from within discord and accepts any valid RegEx for the filter. If you don't want to mess with regex, you can quite literally put a word in there and it'll happily filter just that word out for you. Some features that I want to implement in the future that I haven't yet:

- Customization of deletion messages (thus why the table isn't specific to the DM setting, even though it's the only setting currently)
- Being able to access the logged infractions outside of a eval command or a DB lookup. They're all being tracked at this point, but there's no way of viewing them.


guineawheek:
really nitpicky but the source files are spaced with tabs, while the rest of the project has switched to spaces.

better to fix now while it's not merged

tweirtx:
PyCharm has a feature to automatically convert all indents. Hit CTRL-A to select everything then you should see a convert indents option under the edit menu

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+		await self.check_filters(message)
+
+	async def on_message_edit(self, before, after):
+		await self.check_filters(after)
+
+	"""Commands"""
+
+	@group(invoke_without_command=True)
+	@guild_only()
+	async def filter(self, ctx, advanced = False):
```
endreman0: `advance` should be annotated as a `bool`, so `false`, `no`, etc will be converted to False.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+			results = session.query(WordFilter).filter(WordFilter.guild_id == ctx.guild.id,
+													   WordFilter.enabled == True).all()
+		if results == []:
+			# TODO: Make this a embed
+			await ctx.send("No filters found for this server. Use `{}filter add <>` to add one.".format(
+				ctx.bot.command_prefix))
+			return
+
+		filter_text = ""
+		for filter in results:
```
endreman0: This looks a bit like [potato programming](https://divmod.readthedocs.io/en/latest/philosophy/potato.html), and is a good candidate for a generator expression. Repeated concatenation with `+=` can be replaced with `'\n'.join`. Because the format string depends on the value of `advanced`, making it convoluted to define in one expression, I'd declare it once and then use `map` to format the results like so:

```py
fmt = 'ID {0.id}: `{0.friendly_name}`'
if advanced:
     fmt += ': Pattern: `{0.pattern}`'

filter_text = '\n'.join(map(fmt.format, results))
```

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+				filter_text += ": Pattern: `{}`".format(filter.pattern)
+			if results.index(filter) != (len(results)-1):
+				filter_text += "\n"
+
+		embed = discord.Embed()
+		embed.title = title="Filters for {}".format(ctx.guild.name)
+		embed.add_field(name="Filters", value=filter_text)
+		if ctx.channel.permissions_for(ctx.author).manage_guild:
+			embed.add_field(name="Commands", value="""There are several commands you can use to manage the filters.
+`{0}filter add test` - Adds test as a filter.
```
endreman0: This repeats the example usage. You can access the example usage with `ctx.command.example_usage` and then format in the prefix the same way.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+`{0}filter remove 1` - Removes filter 1
+`{0}filter dm true` - Any messages containing a filtered word will be DMed
+`{0}filter whitelist` - See all of the whitelisted roles
+`{0}filter whitelist add Administrators` - Make the Administrators role whitelisted for the filter.
+`{0}filter whitelist remove Moderators` - Make the Moderators role no longer whitelisted.""".format(self.bot.command_prefix))
+		channel = await self.check_dm_filter(ctx)
+		await channel.send(embed=embed)
+		if isinstance(channel, discord.DMChannel):
```
endreman0: With the above suggestion to return the Member directly, this should check for `discord.Member` instead.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+`{prefix}filter whitelist` - See all of the whitelisted roles
+`{prefix}filter whitelist add Administrators` - Make the Administrators role whitelisted for the filter.
+`{prefix}filter whitelist remove Moderators` - Make the Moderators role no longer whitelisted."""
+
+	@guild_only()
+	@has_permissions(manage_guild=True)
+	@filter.command()
+	async def add(self, ctx, pattern, friendly_name=None):
+		"""Add a pattern to the filter using RegEx. Any word can be added and is tested case-insensitive."""
+		new_filter = WordFilter(guild_id=ctx.guild.id, pattern=pattern, friendly_name=friendly_name or pattern)
```
endreman0: Is `pattern` checked to be a valid regex?

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+			if result is None:
+				await ctx.send("Filter ID {} not found!".format(id))
+				return
+			if result.guild_id != ctx.guild.id:
+				await ctx.send("That Filter does not belong to this guild.")
+				return
+			result.enabled = False
+			await ctx.send("Filter `{}` with name `{}` deleted.".format(result.id,result.friendly_name))
+			self.load_filters(ctx.guild.id)
+	remove.example_usage = "{prefix}filter remove 7"
```
endreman0: The example usage should have the command in \`code formatting\` and include an explanation of what the command does. This also applies to the subcommands below.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+	dm_config.example_usage = "{prefix}filter dm_config True"
+
+	@guild_only()
+	@filter.group(invoke_without_command=True)
+	async def whitelist(self, ctx):
+		"""List all whitelisted roles for this server"""
+		with db.Session() as session:
+			results = session.query(WordFilterRoleWhitelist).filter(WordFilterRoleWhitelist.guild_id == ctx.guild.id).all()
+			roles_text = ""
+			for db_role in results:
```
endreman0: This looks like another good place to use `str.join` rather than repeated concatenation.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+				role = discord.utils.get(ctx.guild.roles, id=db_role.role_id)
```
endreman0: This will be None if the role was deleted.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+	@guild_only()
+	@has_permissions(manage_roles=True)
+	@whitelist.command(name="add")
+	async def whitelist_add(self, ctx, *, role: discord.Role):
+		"""Add a role to the whitelist"""
+		with db.Session() as session:
+			result = session.query(WordFilterRoleWhitelist).filter(WordFilterRoleWhitelist.role_id == role.id).one_or_none()
+			if result is not None:
+				await ctx.send("That role is already whitelisted.")
+			whitelist_entry = WordFilterRoleWhitelist(guild_id=ctx.guild.id, role_id=role.id)
```
endreman0: Adding an already-whitelisted role will display that it is already whitelisted, then still attempt to whitelist it again.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+			session.add(whitelist_entry)
+			await ctx.send("Whitelisted `{}` for this guild.".format(role.name))
```
endreman0: This should be outside the session's `with`-block, so that errors in sending the message (i.e. permissions errors) don't roll back an otherwise successful transaction.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+	@guild_only()
+	@has_permissions(manage_roles=True)
+	@whitelist.command(name="remove")
+	async def whitelist_remove(self, ctx, *, role:discord.Role):
+		"""Remove a role from the whitelist"""
+		with db.Session() as session:
+			result = session.query(WordFilterRoleWhitelist).filter(WordFilterRoleWhitelist.role_id == role.id).one_or_none()
+			if result is None:
+				await ctx.send("That role is not whitelisted.")
+			session.delete(result)
```
endreman0: This will attempt to run even if the result is `None`.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+	setting_type = db.Column(db.String)
+	guild_id = db.Column(db.Integer)
+	value = db.Column(db.String)
+
+
+class WordFilterRoleWhitelist(db.DatabaseObject):
+	__tablename__ = "word_filter_role_whitelist"
+	id = db.Column(db.Integer, primary_key=True)
+	guild_id = db.Column(db.Integer)
+	role_id = db.Column(db.Integer)
```
endreman0: Role IDs are unique and this is essentially a representation of a role in the database, so using the role ID as the primary key looks reasonable to me.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,266 @@
...
+	pattern = db.Column(db.String)
+	infractions = db.relationship("WordFilterInfraction", back_populates="filter")
+
+
+class WordFilterSetting(db.DatabaseObject):
+	__tablename__ = "word_filter_settings"
+	id = db.Column(db.Integer, primary_key=True)
+	setting_type = db.Column(db.String)
+	guild_id = db.Column(db.Integer)
+	value = db.Column(db.String)
```
endreman0: This is a new way of handling settings. Other cogs use one row per guild, one column per setting, with the settings' defaults as column defaults. I'm interested in your reasoning for implementing the filter settings like this.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,256 @@
...
+        with db.Session() as session:
+            results = session.query(WordFilterSetting).filter_by(guild_id=ctx.guild.id, setting_type="dm") \
+                .one_or_none()
+
+            if results is None:
+                results = True
+            else:
+                results = results.value
+
+            if results is "1":
```
endreman0: `"1" is not True`, which looks like an error.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,256 @@
...
+        for id, filter in filters.items():
+            if filter.search(message.content) is not None:
+                await message.channel.send("{}, Banned word detected!".format(message.author.mention), delete_after=5.0)
+                time = datetime.datetime.utcnow()
+                with db.Session() as session:
+                    infraction = WordFilterInfraction(member_id=message.author.id, filter_id=id,
+                                                      timestamp=time,
+                                                      message=message.content)
+                    session.add(infraction)
+                await message.delete()
```
endreman0: If a message matches multiple bans, it will attempt to be deleted multiple times, will raise a `NotFound`.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,256 @@
...
+
+        filter_text = '\n'.join(map(fmt.format, results))
+
+        embed = discord.Embed()
+        embed.title = title = "Filters for {}".format(ctx.guild.name)
+        embed.add_field(name="Filters", value=filter_text)
+        await self.check_dm_filter(ctx, embed)
+
+    filter.example_usage = """`{prefix}filter add test` - Adds test as a filter.
+`{prefix}filter remove 1` - Removes filter 1
```
endreman0: The `example_usage` setter uses `inspect.cleandoc` to fix indentation, so this can be indented one level to be consistent with other commands.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,259 @@
...
+        for id, filter in filters.items():
+            if filter.search(message.content) is not None:
+                await message.channel.send("{}, Banned word detected!".format(message.author.mention), delete_after=5.0)
+                time = datetime.datetime.utcnow()
+                with db.Session() as session:
+                    infraction = WordFilterInfraction(member_id=message.author.id, filter_id=id,
+                                                      timestamp=time,
+                                                      message=message.content)
+                    session.add(infraction)
+                if not deleted:
```
endreman0: This still checks every filter even after a matching one is found. A simpler solution would be to break out of the loop after the message is deleted.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,259 @@
...
+        with db.Session() as session:
+            results = session.query(WordFilterSetting).filter_by(guild_id=ctx.guild.id, setting_type="dm") \
+                .one_or_none()
+
+            if results is None:
+                results = True
+            else:
+                results = results.value
+
+            if results:
```
endreman0: `"0"` is truey, so this condition will pass no matter what is read from the database.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,259 @@
...
+
+    async def check_filters(self, message):
+        if message.author.id == self.bot.user.id:
+            return
+        with db.Session() as session:
+            roles = session.query(WordFilterRoleWhitelist).filter_by(guild_id = message.guild.id).all()
+        whitelisted_ids = set(role.role_id for role in roles)
+        if any(x.id in whitelisted_ids for x in message.author.roles):
+            return
+        filters = {}
```
endreman0: This declaration is never used, as both paths through the try/except below declare `filters`, and should be removable.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,259 @@
...
+
+    dm_config.example_usage = "`{prefix}filter dm_config True` - Makes all messages containining filter lists to be sent through DMs"
+
+    @guild_only()
+    @filter.group(invoke_without_command=True)
+    async def whitelist(self, ctx):
+        """List all whitelisted roles for this server"""
+        with db.Session() as session:
+            results = session.query(WordFilterRoleWhitelist).filter_by(guild_id=ctx.guild.id).all()
+            role_objects = (discord.utils.get(ctx.guild.roles, id=db_role.role_id) for db_role in results)
```
endreman0: This iterates over the guild's entire role list once for every whitelisted role. I would set this up in a large-side-of-reasonable server (say, 150 roles), whitelist 5-10 of them, and see how long this takes.

tweirtx:
[Word filter](https://trello.com/c/avesFrfg/85-word-filter)

dozer/cogs/filter.py
bkeeneykid: This was just me completely forgetting how scopes work, whoops.

dozer/cogs/filter.py
bkeeneykid: The intent that if there's multiple infractions found inside the message that each one is tracked individually (with different filter IDs)

dozer/cogs/filter.py
bkeeneykid: Added 150 roles, whitelisted 10 of them, executed virtually instantly.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,271 @@
...
+        with db.Session() as session:
+            result = session.query(WordFilter).filter_by(id=filter_id).one_or_none()
+            if result is None:
+                await ctx.send("Filter ID {} not found!".format(filter_id))
+                return
+            if result.guild_id != ctx.guild.id:
+                await ctx.send("That Filter does not belong to this guild.")
+                return
+            result.enabled = False
+            await ctx.send("Filter `{}` with name `{}` deleted.".format(result.id, result.friendly_name))
```
endreman0: This and `self.load_filters(ctx.guild.id)` should be outside the session's `with`-block, so that errors while sending the message or reloading filters don't roll back an otherwise successful transaction. Same applies to `dm_config`.

dozer/cogs/filter.py
endreman0: Since #135, finding a role by ID is now O(1) with `Guild.get_role`. It can be virtually instant-er.

dozer/cogs/filter.py
```diff
@@ -0,0 +1,275 @@
...
+
+    @guild_only()
+    @has_permissions(manage_guild=True)
+    @filter.command()
+    async def add(self, ctx, pattern, friendly_name=None):
+        """Add a pattern to the filter using RegEx. Any word can be added and is tested case-insensitive."""
+        try:
+            re.compile(pattern)
+        except re.error as err:
+            await ctx.send("Invalid ReGex! ```{}```".format(err.msg))
```
guineawheek: i don't think this capitalization was intentional

guineawheek: i doubt this spelling was intentional

PR #115, opened 2018-07-05
A few perms changes and modlogs not spamming selfdeafens

tweirtx:


bkeeneykid:
`nomodlog` should probably be changed to not be used as a double negative. I might suggest using something like `create_mod_log` (as to conform to PEP8 variable names too) instead.

dozer/cogs/moderation.py
```diff
@@ -406,15 +407,16 @@ class Moderation(Cog):
             else:
                 user = Deafen(id=ctx.author.id, guild=ctx.guild.id, self_inflicted=True)
                 session.add(user)
+                await ctx.send("Deafening...")
```
endreman0: [Typing notifications](https://discordpy.readthedocs.io/en/rewrite/ext/commands/api.html#discord.ext.commands.Context.typing) would be a better way to show the command is in progress.

dozer/cogs/moderation.py
```diff
@@ -75,11 +75,12 @@ class Moderation(Cog):
         except:
             minutes = 0
         time = (hours * 3600) + (minutes * 60)
-        if time == 0:
+        if time == 0 and modlog:
             await self.mod_log(member=ctx.author, action=punishment.past_participle, target=target, reason=reason, orig_channel=ctx.channel)
         else:
             reasoning = re.sub(pattern=regex_string, string=reason, repl="").lstrip("  ")
-            await self.mod_log(member=ctx.author, action=punishment.past_participle, target=target, reason=reasoning, orig_channel=ctx.channel)
+            if modlog:
```
bkeeneykid: So you've got the `modlog` check in the initial if statement, the you've got another one here... there's got to be a cleaner way of doing this.

endreman0: Neither `reason` nor `reasoning` are used after this point, so you can reassign `reason` here and then do the log after this if suite.

Issue #118, opened 2018-07-20
Dozer doesn't preserve mutes/deafen times across reboots


guineawheek:
title

tweirtx:
I was thinking about that, my thought is to load in the moderation timers from the database when the on_ready event fires

endreman0:
[Implementation of a similar feature](https://github.com/endreman0/Excalibot/blob/26202c0dc6a8ff3c2d033ff45fa174574d1fb279/excalibot/cogs/timer.py#L7-L15) (creating a fake context like that is dirty, but should be avoidable)

tweirtx:
Solved by #122 once it is merged

tweirtx:
Given that this is merged, I am closing this issue

PR #122, opened 2018-07-31
Make timed mutes/deafens restart on bot shutdown

guineawheek:
yeah this patch was pretty involved


tweirtx:
Closes #118 when merged

dozer/cogs/moderation.py
```diff
@@ -627,6 +710,21 @@ class GuildMessageLinks(db.DatabaseObject):
...
+    guild_id = db.Column(db.Integer)
+    actor_id = db.Column(db.Integer)
+    target_id = db.Column(db.Integer)
+    orig_channel_id = db.Column(db.Integer, nullable=True)
+    type = db.Column(db.Integer)
+    reason = db.Column(db.String, nullable=True)
+    target_ts = db.Column(db.Integer)
+
+    type_map = {
+        1: Mute,
```
endreman0: Since you have these as the `type` attribute of the class, you can use that attribute instead of reusing the magic number here. That creates some repetition of the class names, which can be solved by using a dictionary comprehension for this.

guineawheek:
addressed change request and merged branch with head

PR #125, opened 2018-08-15
Fix embeds

tweirtx:


endreman0:
As multiple commands on the production instance are broken, and the fix is extremely simple, this warrants a single review to quickly apply the fix.

PR #128, opened 2018-09-05
Fix the selfdeafen modlog bug

tweirtx:


tweirtx:
[Selfdeafen modlogs](https://trello.com/c/Sem87TA2/88-selfdeafen-modlogs)

PR #129, opened 2018-09-27
Updated &give and &take commands to use embeds.

770352:
Give command:
Added embed_links requirement
Added embed code
Removed old text based message

Take command:
Added embed_links requirement
Added embed code
Removed old text based message


770352:
There we go, The CI build should now be ok.

PR #132, opened 2018-10-04
updated %toa to not actually use toa but work with a more up to date db

guineawheek:
This patch is cherrypicked from my development branch - see #dozer-development for more details

### DIFF
--- a/dozer/__main__.py
+++ b/dozer/__main__.py
@@ -15,6 +15,7 @@ config = {
     'toa': {
         'key': 'Put TOA API key here',
         'app_name': 'Dozer',
+        'teamdata_url': ''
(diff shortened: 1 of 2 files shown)


770352:
Why wouldn't we pull data from straight from TOA's API?

guineawheek:
because TOA's reg data is often hilariously out of date and incomplete

guineawheek:
@770352 because TOA's reg data is often hilariously out of date and incomplete

tweirtx:
Seeing as TOA has fixed your complaint with them and a TOA developer is working on a fix to bring us up to date, I now see no reason for this PR.

PR #133, opened 2018-10-04
added %eval logging

guineawheek:
see title

also cherrypicked from my personal dev/ftc branch


770352:
Just a heads up, Dozer's Offical Prefix is &.

guineawheek:
@770352 I know. I don't care. Plus this commit was cherrypicked from a branch where the Dozers that run it all use `%` as a prefix

PR #136, opened 2018-10-19
Updated TOA command

ofekashery:
A little over a month ago, we updated [our API](https://theorangealliance.org/apidocs).
To see complete details of data accessible via APIv3, please see the [full documentation](https://orange-alliance.github.io/TOA-API).


tweirtx:
@endreman0 please weigh in on this

guineawheek:
i'm p sure dozer will work just fine if you remove `v2` from the end of the toa api url, i think i tested this a long time ago

ofekashery:
> content-type: application/json is implicit if you use await response.json() which is cleaner anyway. previously we did not use that because TOA would return HTML 404 pages (reeeeeee)

The API currently doesn't work without it.

> also you strip away the website handling code; often times team website fields from raw FIRST data lack http[s]:// which is why it was added there in the first place. also TOA for the longest time had zero website data despite the existence of a website field

Make sure that the URLs are filtered and returned correctly **always**.
<br/><br/>
_You are invited_ to check the code in any case of endpoint, it always works.

ofekashery:
The `TOAResponse` didn't help anything

ofekashery:
> TOA still doesn't have data on old teams. People still care about this one. ftc_teams.pickle.gz is what backfilled for them. It's a fallback that gets used less and less but still important to have.

This is exactly what we are doing in the coming days. Within a few days we will have all the teams that have ever been.


> The new TOA url icon may or not be a good choice with the embed's color scheme. This should be up for debate.

I'll look at it

ofekashery:
> I'm going to debate that it's not a good idea, and that TOAResponse is helpful. TOAParser, and TOAResponse with it, hide the details and idiosyncrasies of the TOA API (e.g. requesting a single team returns a list, non-existent teams return an empty list rather than null or a 404 or any other error, and the API will respond with the content type text/html occasionally even though the response body is still JSON) internally and present a cleaner interface for the rest of the code. If the TOA API updates - say, they switch to XML responses, or team_name_short becomes team_nickname akin to TBA, or they switch to returning a single JSON object rather than a list - we can fix it in _toa.py such that we don't have to update the rest of the code at all. Moving the parsing out of the parser relegates the TOAParser to a ratelimited wrapper around an aiohttp client and leaks abstractions about the TOA API (e.g. that JSON is the output format and nonexistent teams return an empty list) into code that shouldn't have to care about them.

If you're not understanding, I'm a TOA Developer. We have no intention of changing the whole API, Certainly not to xml!
Feel free to try, the response always coming back in JSON.
<br/><br/>
>I'm not sure what you're saying here. If TOA makes sure that the URLs are correct, and something has changed since API v2 to make that the case, that's good news. If not, I'd appreciate some clarification.

I don't know what you're talking about, we didn't even have any teams websites until 3 days ago.
All of the teams websites are filtered and we make sure they are standard URLs that include http:// OR https://
<br/><br/>
> The lines are on the thin side and a bit hard to see when shown that small. I have no opinion on the color scheme. I'm more worried about the URL breaking should TOA update their Twitter picture.

We'll do it

endreman0:
> If you're not understanding, I'm a TOA Developer. We have no intention of changing the whole API, Certainly not to xml!
> Feel free to try, the response always coming back in JSON.

That was just an example breaking change. The other two, changing attributes of the returned object and changing the output format from a single-item list to an object, are more relevant. Any breaking change to the TOA API can be contained to just `_toa.py` and not require updating the `toa` command.

> > I'm not sure what you're saying here. If TOA makes sure that the URLs are correct, and something has changed since API v2 to make that the case, that's good news. If not, I'd appreciate some clarification.
> 
> I don't know what you're talking about, we didn't even have any teams websites until 3 days ago.
All of the teams websites are filtered and we make sure they are standard URLs that include http:// OR https://

Good, then once the backup data is unnecessary, the website check will be as well. It sounds like that is still a few days away, though.

> > The lines are on the thin side and a bit hard to see when shown that small. I have no opinion on the color scheme. I'm more worried about the URL breaking should TOA update their Twitter picture.
>
> We'll do it

We meaning TOA developers? We'll do what? Use wider lines in the icon? Update your Twitter PFP? Use a different URL?

ofekashery:
> We meaning TOA developers? We'll do what? Use wider lines in the icon? Update your Twitter PFP? Use a different URL?

We'll use a different URL

ofekashery:
If you have any more questions about the API, please visit [our discord server](https://discord.gg/5fH66UV).

ofekashery:
Now we have all the teams from 2006

endreman0:
Excellent! Since you've already removed the backup data halfway, cleaning it entirely would be a good next step. I'm also still curious about why `TOAResponse` was removed given my comment describing its usefulness.

ofekashery:
I can not find any reason to use the `TOAResponse`

endreman0:
The short answer is it's insurance against changes in TOA's response format. If anything about the response changes (the most likely being fields being renamed), we can add aliases to `TOAResponse` so code outside of `_toa.py` doesn't have to be updated. Say, for instance, `team_key` on `Team` models was renamed to `key`, akin to TheBlueAlliance's equivalent field. With these changes, updating to that change would require updating every usage of the team key; with `TOAResponse`, we just add

```py
# in TOAResponse
@property
def team_key(self):
    return self.key
```
and we're done.

dozer/cogs/toa.py
```diff
@@ -74,17 +74,15 @@ def __init__(self, bot):
...
             await ctx.send("This team does not have any data on it yet, or it does not exist!")
             return
 
-        team_data = json.loads(res)[0]
-
         e = discord.Embed(color=embed_color)
         e.set_author(name='FIRST® Tech Challenge Team {}'.format(team_num),
                      url='https://www.theorangealliance.org/teams/{}'.format(team_num),
-                     icon_url='https://pbs.twimg.com/profile_images/1049159734249623553/SZ34vdcC_400x400.jpg')
+                     icon_url='https://theorangealliance.org/favicon.ico')
```
octocynth: Discord web and Android (at least) don't render `.ico` files properly. Web and shows an empty space where the icon is, and Android shows a dark gray square.

ofekashery: I've changed it already to png

PR #138, opened 2018-10-22
Fixed warns not local modlogging, added modlog DMs

tweirtx:
This is a re-open because the CI stuff was broken


tweirtx:
Never mind it's still broken thanks for breaking GitHub

PR #140, opened 2018-10-22
Fixed warns not local modlogging, added modlog DMs

tweirtx:
I think it might be working again

### DIFF
--- a/dozer/cogs/moderation.py
+++ b/dozer/cogs/moderation.py
@@ -46,7 +46,7 @@ class Moderation(Cog):
         try:
             await target.send(embed=modlog_embed)
         except discord.Forbidden:
-            await orig_channel.send("Attempted to send modlog to user, but they have blocked me.")
+            await orig_channel.send("Failed to DM modlog to user")
         with db.Session() as session:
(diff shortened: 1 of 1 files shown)


dozer/cogs/moderation.py
```diff
@@ -43,7 +43,10 @@ class Moderation(Cog):
         modlog_embed.add_field(name="Requested by", value=f"{actor.mention} ({actor} | {actor.id})", inline=False)
         modlog_embed.add_field(name="Reason", value=reason or "No reason specified", inline=False)
         modlog_embed.add_field(name="Timestamp", value=str(datetime.datetime.now()), inline=False)
-
+        try:
+            await target.send(embed=modlog_embed)
+        except discord.Forbidden:
+            await orig_channel.send("Attempted to send modlog to user, but they have blocked me.")
```
endreman0: Or they no longer share a guild with the bot, or have disabled DMs from guild members... a more generic "failed to DM the user" message would be better here.

tweirtx: Alright, so should I bother with making it run the send before doing the action or should I just change the message?

endreman0: Definitely log before acting, but change the message as well. Even with that fix there are more reasons than just blocking the bot.

PR #144, opened 2018-11-30
Fix style issues discovered by new pylint 2.2 lints

octocynth:


dozer/cogs/_utils.py
```diff
@@ -42,7 +42,6 @@ def example_usage(self, usage):
 
 class Command(CommandMixin, commands.Command):
     """Represents a command"""
-    pass
```
bkeeneykid: Does this not error out due to not having a command body?

PR #148, opened 2019-02-14
Download and re-upload wttr.in images in weather command

octocynth:
Linking to wttr.in in the embed causes Discord to cache the created image, and the weather will change before it expires. Additionally, different regions can cache the image at different times, resulting in multiple users looking at the same message and each seeing a different outdated weather report.

To prevent that, we can download the image from wttr.in ourselves, upload it as an attachment, and make the embed use that attachment as the image. This makes the command slower, but not prohibitively so. Also, because "open original" no longer links to wttr.in, the wttr.in link is set as the embed URL.

### DIFF
--- a/dozer/cogs/tba.py
+++ b/dozer/cogs/tba.py
@@ -1,5 +1,6 @@
 """A series of commands that talk to The Blue Alliance."""
 import datetime
+import io
 import itertools
 import json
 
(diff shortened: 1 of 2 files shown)


tweirtx:
Can someone approve this so we can merge?

PR #151, opened 2019-05-13
billing update to the readme

royalPanic:
A simple update to address that using the GMAPI is no longer free in order to not confuse newcomers looking to set up the bot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,3 @@
-
-
 # Dozer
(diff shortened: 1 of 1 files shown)


octocynth:
The CI failure is due to an unrelated issue. Once the other checks have passed, I'll temporarily disable it to allow merging this.

PR #180, opened 2020-05-19
Round 2 of Issue Closing

tweirtx:
After this PR, I'm backing away for a while to work on something else. Hopefully these issue closures as well as my other PRs have helped restart the heart of the Dozer project :)

### DIFF
--- a/dozer/__main__.py
+++ b/dozer/__main__.py
@@ -17,7 +17,7 @@ config = {
         'key': 'Put TOA API key here',
         'app_name': 'Dozer',
     },
-    'db_url': 'sqlite:///dozer.db',
+    'db_url': 'postgres://POSTGRES URL HERE',
     'gmaps_key': "PUT GOOGLE MAPS API KEY HERE",
     'discord_token': "Put Discord API Token here.",
     'is_backup': False
--- a/dozer/__main__.py
+++ b/dozer/__main__.py
@@ -28,10 +28,12 @@ if os.path.isfile(config_file):
     with open(config_file) as f:
         config.update(json.load(f))
 
-asyncio.get_event_loop().run_until_complete(db_init(config['db_url']))
 
 with open('config.json', 'w') as f:
     json.dump(config, f, indent='\t')
+
+asyncio.get_event_loop().run_until_complete(db_init(config['db_url']))
+
 
 if 'discord_token' not in config:
     sys.exit('Discord token must be supplied in configuration')
--- a/dozer/cogs/moderation.py
+++ b/dozer/cogs/moderation.py
@@ -391,10 +391,10 @@ class Moderation(Cog):
     @command()
     @has_permissions(kick_members=True)
     async def customlog(self, ctx, *, reason):
-        """Sends a message to the mod log specifying the member has been warned without punishment."""
+        """Sends a message to the mod log with custom text."""
         await self.mod_log(actor=ctx.author, action="", target=None, orig_channel=ctx.channel, reason=reason, embed_color=0xFFC400)
 
-    warn.example_usage = """
+    customlog.example_usage = """
     `{prefix}`customlog reason - warns a user for "reason"
     """
 
--- a/dozer/cogs/moderation.py
+++ b/dozer/cogs/moderation.py
@@ -287,6 +287,8 @@ class Moderation(Cog):
     @Cog.listener('on_message_delete')
     async def on_message_delete(self, message):
         """When a message is deleted, log it."""
+        if message.author == self.bot.user:
+            return
         e = discord.Embed(type='rich')
         e.title = 'Message Deletion'
         e.color = 0xFF0000
--- a/dozer/cogs/moderation.py
+++ b/dozer/cogs/moderation.py
@@ -36,22 +36,31 @@ class Moderation(Cog):
 
     """=== Helper functions ==="""
 
-    async def mod_log(self, actor: discord.Member, action: str, target: Union[discord.User, discord.Member], reason, orig_channel=None,
+    async def mod_log(self, actor: discord.Member, action: str, target: Union[discord.User, discord.Member, None], reason, orig_channel=None,
                       embed_color=discord.Color.red(), global_modlog=True):
         """Generates a modlog embed"""
+
+        if target is None:
+            title = "Custom Modlog"
+        else:
+            title = f"User {action}!"
+
         modlog_embed = discord.Embed(
             color=embed_color,
-            title=f"User {action}!"
+            title=title
 
         )
-        modlog_embed.add_field(name=f"{action.capitalize()} user", value=f"{target.mention} ({target} | {target.id})", inline=False)
+        if target is not None:
+            modlog_embed.add_field(name=f"{action.capitalize()} user", value=f"{target.mention} ({target} | {target.id})", inline=False)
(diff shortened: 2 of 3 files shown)


dozer/cogs/moderation.py
```diff
@@ -287,6 +296,8 @@ async def on_message(self, message):
     @Cog.listener('on_message_delete')
     async def on_message_delete(self, message):
         """When a message is deleted, log it."""
+        if message.author == self.bot.user:
```
guineawheek, line 299: Open question: do we only want to not record deletes from Dozer, or generalize to all bots?

dozer/cogs/moderation.py
```diff
@@ -377,6 +388,16 @@ async def warn(self, ctx, member: discord.Member, *, reason):
     `{prefix}`warn @user reason - warns a user for "reason"
     """
 
+    @command()
+    @has_permissions(kick_members=True)
+    async def customlog(self, ctx, *, reason):
+        """Sends a message to the mod log with custom text."""
+        await self.mod_log(actor=ctx.author, action="", target=None, orig_channel=ctx.channel, reason=reason, embed_color=0xFFC400)
+
+    customlog.example_usage = """
```
guineawheek, line 397: I don't question adding customlog, other than "don't we already have a %warn command?" Most mods would probably just write their own "custom" blurbs as regular text messages in a mod log channel.

dozer/cogs/teams.py
```diff
@@ -61,6 +64,49 @@ async def teamsfor(self, ctx, user: discord.Member = None):
...
+                    e = discord.Embed(type='rich')
+                    e.title = 'Members going to {}'.format(event_key)
+                e.add_field(name=f"Team {team}", value=memstr)
+
+        if not found_mems:
+            raise BadArgument("Couldn't find any team members for that event!")
+        else:
+            await ctx.send(embed=e)
+
+    teamsfor.example_usage = """
```
guineawheek, line 106: I think you forgot to make the example usage. 

Also, given that very few FTC events even publicly publish team lists before events, it'll be a while before they even benefit from this one. I remember back in the day, a certain somebody would have a channel that would list all the weekly upcoming events and who was coming to each one - I wonder if we could do something similar with the bot?

dozer/cogs/moderation.py
tweirtx: It was a requested feature ¯\_(ツ)_/¯

dozer/cogs/moderation.py
tweirtx: Recording other bot deletions is probably a good thing

dozer/cogs/teams.py
tweirtx: I did, pushing that now. No idea about the FTC one, but I'm going to get yelled at by someone if I leave it out. Not sure how I feel about the last thing

dozer/cogs/moderation.py
```diff
@@ -36,22 +36,31 @@ def __init__(self, bot):
...
         modlog_embed.add_field(name="Requested by", value=f"{actor.mention} ({actor} | {actor.id})", inline=False)
         modlog_embed.add_field(name="Reason", value=reason or "No reason specified", inline=False)
-        modlog_embed.add_field(name="Timestamp", value=str(datetime.datetime.now()), inline=False)
+        modlog_embed.timestamp = datetime.datetime.utcnow()
         try:
             await target.send(embed=modlog_embed)
         except discord.Forbidden:
             await orig_channel.send("Failed to DM modlog to user")
+        except AttributeError:
+            pass
```
octocynth, line 63: This could mask actual bugs in the code if something other than a User/Member is provided. This should be handled with a check if `target is None` before sending the DM.

dozer/cogs/teams.py
```diff
@@ -61,6 +64,49 @@ async def teamsfor(self, ctx, user: discord.Member = None):
...
+    @command()
+    @guild_only()
+    async def compcheck(self, ctx, event_type: str, event_key):
+        """Allows you to see people in the Discord server that are going to a certain competition."""
+        if event_type.lower() == "frc":
+            teams_raw = await TBA(ctx.bot).session.event_teams(event_key)
+            teams = [team.team_number for team in teams_raw]
+        elif event_type.lower() == "ftc":
+            teams_raw = json.loads(await TOA(ctx.bot).parser.req(f"/api/event/{event_key}/teams"))
+            teams = [team['team']['team_number'] for team in teams_raw]
```
octocynth, line 76: Creating a new `TBA`/`TOA` cog every time this command is called is likely going to cause bugs in the future. Dozer cogs assume they're only instantiated once. You can find the existing TBA/TOA cog with the [`Bot.get_cog`](https://discordpy.readthedocs.io/en/latest/ext/commands/api.html#discord.ext.commands.Bot.get_cog) method.

dozer/cogs/teams.py
```diff
@@ -61,6 +64,49 @@ async def teamsfor(self, ctx, user: discord.Member = None):
...
+                    if len(newmemstr + memstr) > 1023:
+                        e.add_field(name=f"Team {team}", value=memstr)
+                        memstr = ""
+                    memstr += newmemstr
+                    found_mems = True
+            if len(memstr) > 0:
+                if len(e.fields) == 25:
+                    await ctx.send(embed=e)
+                    e = discord.Embed(type='rich')
+                    e.title = 'Members going to {}'.format(event_key)
```
octocynth, line 98: This is likely to cause spam when used with i.e. a championship division or DCMP. This would be a good place for pagination.

dozer/__main__.py
```diff
@@ -17,7 +17,7 @@
         'key': 'Put TOA API key here',
         'app_name': 'Dozer',
     },
-    'db_url': 'sqlite:///dozer.db',
+    'db_url': 'postgres://POSTGRES URL HERE',
```
bkeeneykid, line 20: While this does technically fix the problem, the problem is that if there's any error in our `db_init` script, the config file will still fail to generate at all since it runs before the file is saved. See line 33 in `__main__.py`.

dozer/cogs/teams.py
tweirtx: Oops, didn't know that

dozer/cogs/teams.py
tweirtx: I don't know how to paginate

dozer/cogs/moderation.py
tweirtx: Ok

dozer/cogs/teams.py
```diff
@@ -61,6 +62,56 @@ async def teamsfor(self, ctx, user: discord.Member = None):
...
+    @bot_has_permissions(add_reactions=True, embed_links=True,
+                         read_message_history=True)
+    async def compcheck(self, ctx, event_type: str, event_key):
+        """Allows you to see people in the Discord server that are going to a certain competition."""
+        if event_type.lower() == "frc":
+            teams_raw = await ctx.bot.get_cog("TBA").session.event_teams(event_key)
+            teams = [team.team_number for team in teams_raw]
+        elif event_type.lower() == "ftc":
+            teams_raw = json.loads(await ctx.bot.get_cog("TOA").parser.req(f"/api/event/{event_key}/teams"))
+            teams = [team['team']['team_number'] for team in teams_raw]
```
octocynth, line 76: This will raise an exception if the event doesn't exist.

PR #181, opened 2020-06-01
Add News Functionality

bkeeneykid:
Adds ability to subscribe to a number of RSS and other feeds.


dozer/cogs/news.py
```diff
@@ -0,0 +1,451 @@
...
+                raise BadArgument(f"There is already a subscription of {source.full_name} with data {data} "
+                                  f"in {channel.mention}")
+
+            added = await source.add_data(data_obj)
+            if not added:
+                DOZER_LOGGER.error(f"Failed to add data {data_obj} to source {source.full_name}")
+                await ctx.send("Failed to add new data source. Please contact the Dozer Administrators.")
+                return
+
+            # TODO: Check if source is already added before adding
```
tweirtx, line 207: Did you mean to leave this here?

dozer/cogs/news.py
```diff
@@ -0,0 +1,451 @@
...
+                               f"{data_obj} found.")
+                return
+            elif len(sub) > 1:
+                DOZER_LOGGER.error(f"More that one subscription of {source.full_name} for channel "
+                                   f"{channel.mention} with data {data} found when attempting to delete.")
+                await ctx.send(f"More that one subscription of {source.full_name} for channel {channel.mention} "
+                               f"with data {data} found. Please contact the Dozer administrator for help.")
+                return
+
+            # TODO: Check if any other subscriptions use this data first
```
tweirtx, line 269: What about this one?

dozer/sources/RedditSource.py
```diff
@@ -0,0 +1,229 @@
...
+
+        expiry_seconds = response['expires_in']
+        time_delta = datetime.timedelta(seconds=expiry_seconds)
+        self.expiry_time = datetime.datetime.now() + time_delta
+
+    async def request(self, url, *args, headers=None, **kwargs):
+        """Make a request using OAuth2 (or not, if it's been disabled)"""
+        if datetime.datetime.now() > self.expiry_time:
+            DOZER_LOGGER.info("Refreshing Reddit token due to expiry time")
+            await self.get_token()
```
octocynth, line 67: Should this be skipped if OAuth is disabled?

dozer/sources/RedditSource.py
```diff
@@ -0,0 +1,229 @@
...
+        else:
+            url = f"{self.api_url}/{url}"
+
+        response = await self.http_session.get(url, headers=headers, *args, **kwargs)
+
+        if response.status == 401:
+            if 'www-authenticate' in response.headers:
+                DOZER_LOGGER.info("Reddit token expired when request made, requesting new token and retrying.")
+                await self.get_token()
+                return await self.request(url, headers, *args, **kwargs)
```
octocynth, line 85: This passes `headers` as the first argument of `*args`. Is that intentional? The same applies to `TwitchSource`.

dozer/sources/RedditSource.py
```diff
@@ -0,0 +1,229 @@
...
+            elif subreddit_about['kind'] == "Listing":
+                # Reddit "helpfully" redirected us to subreddit search. Let's get some helful error messages out of
+                # this...
+                subreddits = [subreddit['data']['display_name'] for subreddit in subreddit_about['data']['children']
+                              if not subreddit['data']['over18']]
+                if len(subreddits) == 0:
+                    raise DataBasedSource.InvalidDataException(f"No subreddit found for search string {text}")
+                elif len(subreddits) > 1:
+                    raise DataBasedSource.InvalidDataException(f"No exact match was found, but multiple similar "
+                                                               f"subredidts found. Did you mean any of the following:"
```
octocynth, line 129: s/subredidts/subreddits

dozer/sources/RedditSource.py
```diff
@@ -0,0 +1,229 @@
...
+                self.seen_posts.add(post['data']['name'])
+                if first_time:
+                    continue
+
+                embed = self.generate_embed(post['data'])
+                plain = self.generate_plain_text(post['data'])
+                posts[post['data']['subreddit']] = {
+                    'embed': [embed],
+                    'plain': [plain]
+                }
```
octocynth, line 193: This will only show one post for each news run.

dozer/sources/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""A collection of sources for various FIRST-related news sources"""
+
+from .RSSSources import *
+from .TwitchSource import TwitchSource
+from .RedditSource import RedditSource
+from .AbstractSources import Source, DataBasedSource
+
+sources = [TwitchSource, RedditSource]
+sources += [source for source in RSSSource.__subclasses__() if not source.disabled]
```
octocynth, line 9: This ignores `TwitchSource.disabled` and `RedditSource.disabled`.

dozer/sources/TwitchSource.py
```diff
@@ -0,0 +1,208 @@
...
+    async def clean_data(self, text):
+        """Request user data from Twitch to verify the username exists and clean the data"""
+        try:
+            user_obj = self.users[text]
+        except KeyError:
+            json = await self.request('users', params={'login': text})
+            if len(json['data']) == 0:
+                raise DataBasedSource.InvalidDataException("No user with that login name found")
+            elif len(json['data']) > 1:
+                raise DataBasedSource.InvalidDataException("More than one user with that login name found")
```
octocynth, line 110: In this case, what can the user do to indicate which streamer they want?

dozer/sources/TwitchSource.py
```diff
@@ -0,0 +1,208 @@
...
+                    'embed': [embed],
+                    'plain': [plain]
+                }
+
+                self.seen_streams.add(stream['id'])
+
+        return posts
+
+    def generate_embed(self, data, games):
+        """Given data on a stream and a dict of games, assemble a embed"""
```
octocynth, line 173: s/a embed/an embed/

dozer/cogs/news.py
```diff
@@ -0,0 +1,454 @@
...
+                    subs = await NewsSubscription.get_by(source=source.short_name)
+                    data = {sub.data for sub in subs}
+                    await self.sources[source.short_name].first_run(data)
+                else:
+                    await self.sources[source.short_name].first_run()
+            except ElementTree.ParseError as err:
+                del self.sources[source.short_name]
+                DOZER_LOGGER.error(f"Parsing error in source {source.short_name}: {err}")
+        self.get_new_posts.change_interval(minutes=self.bot.config['news']['check_interval'])
+        self.get_new_posts.start()
```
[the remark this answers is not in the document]
bkeeneykid: Ah, I was thinking that this would be blocking to bot startup, but it looks like that's not true.

dozer/sources/RSSSources.py
```diff
@@ -0,0 +1,240 @@
...
+    """Blog posts by John V Neun, 148 Head Engineer"""
+    url = "https://johnvneun.com/blog?format=rss"
+    base_url = "https://johnvneun.com/"
+    full_name = "JVN's Blog"
+    short_name = "jvn"
+    aliases = ['148', 'robowranglers']
+    description = "Blog posts by John V Neun, 148 Head Engineer"
+    color = discord.colour.Color(value=000000)
+    disabled = True
+    # For some reason, the XML parser hates this source. Investigate later.
```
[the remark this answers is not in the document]
bkeeneykid: It looks like since JVN's blog uses HTTP/2.0, the keep-alive parameter is required, but when it's added, it only recieves part of the response, and since the RSS for the site is fairly large, this fails to parse it (invalid XML). It looks like the issue is with aiohttp instead of the XML parser. I still don't have a fix.

dozer/sources/__init__.py
bkeeneykid: That's intentional. For these larger, data based classes I feel like disabling them in such way (at least at bot startup) shouldn't be as necessary. For these, just remove it from the list above.

dozer/sources/TwitchSource.py
bkeeneykid: I have no clue. I have no idea if this is even possible. I added error handling just in case.

PR #185, opened 2020-06-20
Fix multiple bugs in News cog

bkeeneykid:


bkeeneykid:
Travis CI is a bit loaded right now, so the CI might take a bit to pass, but I think this PR should be ready to merge.

I do want to do some more robust error handling in the future but for now this seems stable enough.

tweirtx (dismissed):

requirements.txt
```diff
@@ -1,5 +1,5 @@
 asyncpg
-discord.py[voice]
+discord.py[voice]==1.4.1
```
octocynth, line 2: Right now this will cause problems with the intents code in `dozer/__main__.py`, as that only exists on 1.5+. Maybe this should be a `>=` constraint?

octocynth (changes requested):

requirements.txt
tweirtx: Or at least a version bump? We probably ought to do that anyway

tweirtx (approved):

octocynth (approved):

PR #194, opened 2020-09-21
Fix error on timeout when no member role set

tweirtx:


bkeeneykid (approved):

octocynth (changes requested):
Defaulting to using the default role will prevent `&timeout` from working if there is a role (e.g. the actual member role) that grants send permissions on the channel. A moderator shouldn't have to fuss with settings when they'd rather make a channel please be quiet right now thank you, which is why the default case is to mute everything instead of emitting an error.

The real solution is to fix #195, but it looks like the `MemberRole` this inserts isn't ever used; `&memberconfig` will insert the row if it's missing. The faster fix, then, is to not insert here at all. As long as `settings` exists and `settings.member_role is None` we'll get the desired behavior.

PR #196, opened 2020-10-10
Completed TempGive command for dozer

AidanFromProgramming:


tweirtx (approved):

guineawheek (approved):
seems fine, although shouldn't we also use the presences intent? but imo that's for another PR

PR #197, opened 2020-10-12
fix jvn's blog

guineawheek:
turns out you need a user-agent otherwise his host will 403 you causing the xml errors

making a pr because jvn's blog is more useful for frc kids anyway

bottom text


guineawheek:
(side note: it would be nice if the news system was able to fail more gracefully -- if one of the rss sources 403s on startup there's a good chance the rest of the sources won't load either. and that's bad)

tweirtx (approved):

octocynth (approved):

PR #198, opened 2020-10-17
Completed Auto Push-To-Talk Issue 

AidanFromProgramming:


dozer/cogs/voice.py
```diff
@@ -124,6 +183,45 @@ async def get_by(cls, **kwargs):
...
+
+    __uniques__ = 'id'
+
+    @classmethod
+    async def initial_create(cls):
+        """Create the table in the database"""
+        async with db.Pool.acquire() as conn:
+            await conn.execute(f"""
+            CREATE TABLE {cls.__tablename__} (
+            id SERIAL PRIMARY KEY NOT NULL,
```
octocynth, line 198: What's your thoughts on using the channel ID as the primary key? That should speed up the query in `on_PTT_check` and simplify the `autoptt` command.

dozer/cogs/voice.py
```diff
@@ -27,6 +28,64 @@ async def on_voice_state_update(self, member, before, after):
...
+            await voice_channel.set_permissions(target=everyone, overwrite=perms)
+
+    @command()
+    @bot_has_permissions(manage_channels=True)
+    @has_permissions(manage_channels=True)
+    async def autoptt(self, ctx, voice_channel: discord.VoiceChannel, ptt_threshold: int):
+        """Configures AutoPtt limit for when members join/leave voice channels ptt is enabled"""
+        config = await AutoPTT.get_by(channel_id=voice_channel.id)
+        if len(config) != 0:
+            await AutoPTT.delete(id=config[0].id)
```
octocynth, line 66: If auto-PTT is set for a channel and then set to 0/a negative number, the previous config would be erased. This should be moved below the following block. Or, if the channel ID is used as the primary key (see further down), this could be removed and `update_or_add` will take care of the `UPDATE` if needed.

dozer/cogs/voice.py
```diff
@@ -27,6 +28,64 @@ async def on_voice_state_update(self, member, before, after):
...
+            if before.channel is not None:
+                # leave event, take role
+                voice_channel = before.channel
+                total_users = len(before.channel.members)
+                config = await AutoPTT.get_by(channel_id=before.channel.id)
+            if after.channel is not None:
+                # join event, give role
+                voice_channel = after.channel
+                total_users = len(after.channel.members)
+                config = await AutoPTT.get_by(channel_id=after.channel.id)
```
octocynth, line 48: If a user moves between voice channels, both `before.channel` and `after.channel` will be non-null. It looks like PTT won't be updated on `before.channel` in that case.

dozer/cogs/voice.py
```diff
@@ -27,6 +28,64 @@ async def on_voice_state_update(self, member, before, after):
...
+
+            everyone = voice_channel.guild.default_role  # grab the @everyone role
+            perms = voice_channel.overwrites_for(everyone)  # grab the @everyone overwrites
+
+            if total_users > config[0].ptt_limit:
```
octocynth, line 53: Does this handle the case where the channel hasn't had auto-PTT configured, and `config` is empty?

octocynth (changes requested):
Thanks for the PR! 💜 Aside from the inline comments, it would be great to have a way to disable auto-PTT for a channel once it has been enabled.

dozer/cogs/voice.py
```diff
@@ -1,15 +1,30 @@
...
+from .info import blurple
 from .. import db
 
 
 class Voice(Cog):
     """Commands interacting with voice."""
 
-    @Cog.listener('on_voice_state_update')
+    @staticmethod
+    async def autoPTTCheck(voice_channel):
```
guineawheek, line 14: incredibly minor nitpick but shouldn't we use snake_case here

dozer/cogs/voice.py
```diff
@@ -27,6 +42,60 @@ async def on_voice_state_update(self, member, before, after):
...
+            else:
+                e.add_field(name='Error', value='AutoPTT has not been configured for voice channel "**{}**"'
+                            .format(voice_channel))
+        else:
+            ent = AutoPTT(
+                channel_id=voice_channel.id,
+                ptt_limit=ptt_threshold
+            )
+
+            await ent.update_or_add()
```
octocynth, line 87: With an autoincrementing primary key, this will always insert a new record instead of updating an existing one. Since the order records are returned from a `SELECT` statement isn't specified unless there's an `ORDER BY` clause, `auto_ptt_check` could fetch any record from the channel's history, which I'm pretty confident isn't the intent. If `channel_id` is the primary key for the `AutoPTT` table, this will work as expected.

octocynth (changes requested):

octocynth (approved):

guineawheek (approved):

PR #202, opened 2020-11-02
Dozer Levels

### LINKED ISSUE #158: Levels

AidanFromProgramming:
I think I am at the point where I just need y'all to review my code and make sure nothing has been missed

close #158


dozer/cogs/levels.py
```diff
@@ -0,0 +1,584 @@
...
+            lvl += 1
+        return lvl - 1
+
+    async def check_new_roles(self, guild, member, cached_member):
+        """Check and see if a member has qualified to get a new role"""
+        roles = self._level_roles.get(guild.id)
+        for level_role in roles:
+            role = guild.get_role(level_role.role_id)
+            if self.level_for_total_xp(cached_member.total_xp) >= level_role.level and role not in member.roles:
+                await member.add_roles(role)
```
octocynth, line 143: This makes one request per role, which will be ratelimited pretty heavily. This could collect the list of roles that need to be added in this loop and then add them all at once afterward.

dozer/cogs/levels.py
```diff
@@ -0,0 +1,584 @@
...
+        db_record = await db.Pool.fetchrow(f"""
+            WITH ranked_xp AS (
+                SELECT user_id, rank() OVER (ORDER BY total_xp DESC) FROM {MemberXP.__tablename__}
+                WHERE guild_id = $1
+            ) SELECT rank FROM ranked_xp WHERE user_id = $2;
+        """, ctx.guild.id, member.id)
+
+        total_xp = cache_record.total_xp
+
+        count = ctx.guild.member_count
```
octocynth, line 387: It would be more accurate to fetch this from the database as number of records with this guild ID. As-is, it's possible to show e.g. "# 100 of 90" if enough high-ranked users have left the server.

dozer/cogs/levels.py
```diff
@@ -0,0 +1,584 @@
...
+    @command(aliases=["configurelevels"])
+    @guild_only()
+    @has_permissions(administrator=True)
+    async def configureranks(self, ctx):
+        """Configure dozer ranks:tm: any key word arguments not entered will be treated as default"""
+        args = self.kwarg_parsing(ctx.message.content)  # Parse for kwargs
+
+        xp_min = max(min(int(args.get("min")), 32767), 0) if args.get("min") else 5
+        xp_max = max(min(int(args.get("max")), 32767), 0) if args.get("max") else 15
+        xp_cooldown = max(min(int(args.get("cooldown")), 32767), 0) if args.get("cooldown") else 15
```
octocynth, line 301: Clamping the value into range is likely not what the user expected or wanted. If these values are out of bounds, the command should error.

octocynth (changes requested):

dozer/cogs/levels.py
```diff
@@ -0,0 +1,584 @@
...
+
+        await ctx.send("Done")
+
+    meesyncs.example_usage = """
+        `{prefix}meesyncs`: Sync ranking data from the mee6 API to dozer's database
+        """
+
+    async def update_server_settings_cache(self):
+        """Updates the server settings cache from the database"""
+        await asyncio.sleep(0.75)  # Give the database a few to update
```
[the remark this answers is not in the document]
AidanFromProgramming: Not sure, but it doesn't work without it

dozer/cogs/levels.py
AidanFromProgramming: I added a mix of both uses the server size when the database only doesn't have the full server, and a the database total when the database is bigger than total members

tweirtx (approved):

dozer/cogs/levels.py
```diff
@@ -52,24 +53,25 @@ async def meesyncs(self, ctx):
...
-                        last_given_at=ctx.message.created_at.replace(tzinfo=timezone.utc)
-                    )
-                    await ent.update_or_add()
-                if page % 2:
-                    await msg.edit(content=progress_template.format(page=page))
-            else:
-                break
-            time.sleep(1.25)  # Slow down api calls as to not anger cloudflare
+        for page in itertools.count():
+            async with aiohttp.ClientSession() as session:
```
octocynth, line 57: `ClientSession`s carry a lot of state and are meant to be reused for many requests. A single session should be created when the cog is created and reused between calls. For reference, see how [the `News` cog](https://github.com/FRCDiscord/Dozer/blob/b467ffd70d196aef48ec38c5ca9daddcd0c5fb2f/dozer/cogs/news.py) creates and uses its `http_session` attribute.

dozer/cogs/levels.py
```diff
@@ -252,9 +248,31 @@ def _ensure_sync_running(self):
...
-    @has_permissions(administrator=True)
+    async def checkrolelevels(self, ctx):
+        """Displays all level associated roles"""
+        roles = sorted(self._level_roles.get(ctx.guild.id), key=lambda entry: entry.level)  # Sort roles based on level
+        e = discord.Embed(title=f"Level roles for {ctx.guild}", color=blurple)
+        if roles:
+            e.description = f"This server has {len(roles)} level roles"
+            for level_role in roles:
+                role = ctx.guild.get_role(level_role.role_id)
+                if_unavailable = f"<@{level_role.role_id}>"
```
octocynth, line 261: If the role doesn't exit, its mention isn't easy for a user to understand. This message should be that the role has been deleted or something similar.

dozer/cogs/levels.py
```diff
@@ -284,86 +302,129 @@ async def setrolelevel(self, ctx, role: discord.Role, level: int):
...
+        """Set the range of a servers levels random xp"""
         if xp_min > xp_max:
             raise BadArgument("XP_min cannot be greater than XP_max!")
-
+        await self._config_guild_setting(ctx, xp_min=xp_min, xp_max=xp_max)
+
+    @configureranks.command()
+    @has_permissions(manage_guild=True)
+    async def setcooldown(self, ctx, cooldown: int):
+        """Set the time between messages before xp is calculated again"""
```
octocynth, line 347: Is `cooldown` a number of minutes or seconds? The help message should say which one.

dozer/cogs/levels.py
```diff
@@ -284,86 +302,129 @@ async def setrolelevel(self, ctx, role: discord.Role, level: int):
...
-        await ctx.send(embed=embed)
+            enabled = "Enabled" if settings.enabled else "Disabled"
+            embed.set_author(name=ctx.guild, icon_url=ctx.guild.icon_url)
+            embed.add_field(name=f"Levels are {enabled} for {ctx.guild}", value=f"XP min: {settings.xp_min}\n"
+                                                                                f"XP max: {settings.xp_max}\n"
+                                                                                f"Cooldown: {settings.xp_cooldown} Seconds\n"
+                                                                                f"Notification channel: {notify_channel}")
+            await ctx.send(embed=embed)
+        else:
+            await ctx.send("Levels not configured for this server")
```
octocynth: This should be merged into the `configureranks` command. Settings commands usually serve to both get and set settings, for example the `&starboard` command.

octocynth (changes requested):

dozer/cogs/levels.py
open-sourceress: Hmm, I don't think this got resolved. Instead of telling the user an invalid command was called, `&configureranks` should show the current configuration, instead of having a separate `&getlevelsconfig` command.

open-sourceress (changes requested):
Looks good! After https://github.com/FRCDiscord/Dozer/pull/202#discussion_r522449831 is resolved this should be ready to merge.

tweirtx (dismissed):

open-sourceress (dismissed):
🎉

tweirtx (dismissed):

bkeeneykid (approved):

open-sourceress (approved):

PR #203, opened 2020-11-02
New embeds for message logs

AidanFromProgramming:
I spent some time formatting the embeds for message logs, and added a method for handing raw api payloads in the event a message wasn't in the cache. I also increased the message cache size from 1000 to 5000 which is still quite low for a discord bot of dozers scale

### DIFF
--- a/dozer/cogs/moderation.py
+++ b/dozer/cogs/moderation.py
@@ -366,7 +366,7 @@ class Moderation(Cog):
         channel_id = str(payload.channel_id)
         user_id = str(author['id'])
         if (self.bot.get_user(int(author['id']))).bot:
-            return
+            return  # Breakout if the user is a bot
         message_id = str(payload.message_id)
         link = f"https://discordapp.com/channels/{str(guild_id)}/{str(channel_id)}/{str(message_id)}"
         mention = f"<@!{user_id}>"
@@ -392,7 +392,7 @@ class Moderation(Cog):
     async def on_message_edit(self, before, after):
         """Logs message edits."""
         await self.check_links(after)
-        if before.author.bot:
+        if before.author.bot or after.author.bot:
             return
         if after.edited_at is not None or before.edited_at is not None:
             # There is a reason for this. That reason is that otherwise, an infinite spam loop occurs
--- a/dozer/cogs/moderation.py
+++ b/dozer/cogs/moderation.py
@@ -308,10 +308,10 @@ class Moderation(Cog):
         message_id = int(payload.message_id)
         message_created = discord.Object(message_id).created_at
         embed = discord.Embed(title="Message Deleted",
-                              description="Message Deleted In: " + str(message_channel.mention),
+                              description=f"Message Deleted In: {message_channel.mention}",
                               color=0xFF00F0, timestamp=message_created)
         embed.add_field(name="Message", value="N/A", inline=False)
-        embed.set_footer(text=f"MessageID: {str(message_id)}; Sent at")
+        embed.set_footer(text=f"MessageID: {message_id}; Sent at")
         message_log_channel = await self.edit_delete_config.query_one(guild_id=guild.id)
         if message_log_channel is not None:
             channel = guild.get_channel(message_log_channel.messagelog_channel)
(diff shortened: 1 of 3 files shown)


dozer/cogs/moderation.py
```diff
@@ -292,88 +298,133 @@ async def on_message(self, message):
...
-                for x in i.fields:
-                    e.add_field(name=x.name, value=x.value)
-                e.add_field(name="Footer", value=i.footer)
+        audit = await self.check_audit(message.guild)
+        embed = discord.Embed(title="Message Deleted",
+                              description=f"Message Deleted In: {str(message.channel.mention)}\nSent by: {str(message.author.mention)}",
+                              color=0xFF0000, timestamp=message.created_at)
+        embed.set_thumbnail(url=message.author.avatar_url)
+        if audit:
+            if str(audit.target) == str(message.author.name + "#" + message.author.discriminator):
```
octocynth, line 332: What's the reasoning for comparing equality by `str(...)`? Does `audit.target == message.author` not do the expected thing here?

dozer/cogs/moderation.py
```diff
@@ -292,88 +298,133 @@ async def on_message(self, message):
...
+                auditMember = await message.guild.fetch_member(audit.user.id)
+                embed.add_field(name="Message Deleted By: ", value=str(auditMember.mention), inline=False)
```
octocynth, line 334: It's probably worth including the user's username\#discriminator and ID here too. When a user leaves the guild or the client hasn't loaded the full member list, mentions are frustratingly opaque.

dozer/cogs/moderation.py
```diff
@@ -292,88 +298,133 @@ async def on_message(self, message):
...
+                await channel.send(embed=embed)
+
+    @Cog.listener()
+    async def on_raw_message_edit(self, payload):
+        """Logs message edits that are not currently in the bots message cache"""
+        if payload.cached_message:
+            return
+        Mchannel = self.bot.get_channel(int(payload.channel_id))
+        guild = Mchannel.guild
+        print(payload)
```
octocynth, line 357: Looks like this is a leftover from debugging that can be removed.

dozer/cogs/moderation.py
```diff
@@ -292,88 +298,133 @@ async def on_message(self, message):
...
+        try:
+            content = payload.data['content']
+        except KeyError:
+            content = None
+        try:
+            author = payload.data['author']
+        except KeyError:
+            return
```
octocynth, line 365: When does `payload.data['author']` not exist?

dozer/cogs/moderation.py
```diff
@@ -292,88 +298,133 @@ async def on_message(self, message):
...
+        mention = f"<@!{user_id}>"
+        avatar_link = f"http://cdn.discordapp.com/avatars/{str(user_id)}/{author['avatar']}.webp?size=1024"
+        embed = discord.Embed(title="Message Edited",
+                              description=f"[MESSAGE]({link}) From {mention}\nEdited In: {str(Mchannel.mention)}", color=0xFFC400)
+        embed.set_thumbnail(url=avatar_link)
+        embed.add_field(name="Original", value="N/A", inline=False)
+        if content:
+            embed.add_field(name="Edited", value=str(content[0:1023]), inline=False)
+            if len(content) > 1024:
+                embed.add_field(name="Edited Continued", value=str(content[1024:2000]), inline=False)
```
octocynth, line 382: Are the `str(...)` calls here necessary? F-strings format expressions with `str(...)` internally already.

dozer/cogs/moderation.py
```diff
@@ -292,88 +298,133 @@ async def on_message(self, message):
...
+            channel = guild.get_channel(message_log_channel.messagelog_channel)
+            if channel is not None:
+                await channel.send(embed=embed)
 
     @Cog.listener('on_message_edit')
     async def on_message_edit(self, before, after):
         """Logs message edits."""
         await self.check_links(after)
-        if before.author.bot:
+        if before.author.bot or after.author.bot:
```
octocynth, line 396: Is there a case where `before.author.bot` and `after.author.bot` differ? AFAIK it's impossible to convert between a user and bot account.

dozer/cogs/moderation.py
```diff
@@ -292,88 +298,133 @@ async def on_message(self, message):
...
+                                              f"\nEdited In: {str(before.channel.mention)}", color=0xFFC400,
+                                  timestamp=after.edited_at)
+            embed.set_thumbnail(url=after.author.avatar_url)
+            if before.content:
+                embed.add_field(name="Original", value=str(before.content[0:1023]), inline=False)
+                if len(before.content) > 1024:
+                    embed.add_field(name="Original Continued", value=str(before.content[1024:2000]), inline=False)
+                embed.add_field(name="Edited", value=str(after.content[0:1023]), inline=False)
+                if len(after.content) > 1024:
+                    embed.add_field(name="Edited Continued", value=str(after.content[1024:2000]), inline=False)
```
octocynth, line 416: Same as  my comments on L366-L382, these `str(...)` calls look like they could be removed.

octocynth (changes requested):

dozer/cogs/moderation.py
AidanFromProgramming: Never...

dozer/cogs/moderation.py
guineawheek: This was actually the original design rationale for having the author and author pingable be separate fields.

dozer/cogs/moderation.py
octocynth: In that case it's probably best to let `payload.data['author']` raise its error so that we get an error report instead of failing silently if that ever changes.

octocynth (changes requested):
Looks good! One more change and then this should be good to merge.

tweirtx (approved):

octocynth (approved):

PR #211, opened 2020-11-06
Added noreturn option to eval command

AidanFromProgramming:
When noreturn is specified no evaluation embed is sent and the code is run


guineawheek:
Is this necessary?

1). I can't think of a time when I don't want the bot to send an embed in an eval, if the embed sent is too big to be sent, that's usually an error on my part, and it helps to have confirmation that the code actually _ran_

2). making subcommands out of eval, which is supposed to be able to run arbitrary python code, is screwy.

guineawheek:
It seems the primary use case is for using Dozer to say things.

In that case, why not just make a dev-only `&say` command?

tweirtx (approved):

PR #215, opened 2020-11-09
Added catch for when bot isn't able to read audit log 

AidanFromProgramming:
Accidentally closed the first pull request

### DIFF
--- a/dozer/cogs/moderation.py
+++ b/dozer/cogs/moderation.py
@@ -7,6 +7,7 @@ from typing import Union
 from logging import getLogger
 
 import discord
+from discord import Forbidden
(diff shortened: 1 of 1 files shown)


tweirtx (approved):

dozer/cogs/moderation.py
```diff
@@ -7,6 +7,7 @@
 from logging import getLogger
 
 import discord
+from discord import Forbidden
```
guineawheek, line 10: why not use `discord.Forbidden` instead of a specific import?

AidanFromProgramming: Because pycharm automatically imported it, nothing in particular guided that decision

octocynth (approved):

PR #218, opened 2020-11-10
Fix starboard manual addition command

bkeeneykid:


bkeeneykid:
So the solution here is really hacky and will break as soon as anybody reacts with the starboard emoji, then removing it. The better solution would likely be to add an extra row in the DB for manual additions and don't worry about the emoji count for those.... but changing the DB schema at this point would be a pain, and I'm okay with this solution instead.

tweirtx (approved):

tweirtx (approved):

octocynth (approved):

PR #219, opened 2020-11-11
Dynamic prefixes

AidanFromProgramming:
Added the ability to configure per guild dynamic prefixes to prevent prefix conflicts with other bots, and to allow customization based on guild owner needs

### DIFF
--- a/dozer/cogs/_utils.py
+++ b/dozer/cogs/_utils.py
@@ -3,9 +3,7 @@ import asyncio
 import inspect
 import logging
 import typing
-from asyncore import loop
 from collections.abc import Mapping
-from concurrent.futures.thread import ThreadPoolExecutor
 
 import discord
 from discord.ext import commands
--- a/dozer/cogs/_utils.py
+++ b/dozer/cogs/_utils.py
@@ -313,7 +313,7 @@ class PrefixHandler:
         self.prefix_cache = {}
 
     def handler(self, bot, message):
-        """Process the dynamic prefix from message"""
+        """Process the dynamic prefix for each message"""
         dynamic = self.prefix_cache.get(message.guild.id)
         return dynamic if dynamic else self.default_prefix
 
@@ -325,9 +325,8 @@ class PrefixHandler:
         logger.info(f"{len(prefixes)} prefixes loaded from database")
 
 
-
 class DynamicPrefixEntry(db.DatabaseTable):
-    """Holds the roles of those who leave"""
+    """Holds the custom prefixes for guilds"""
     __tablename__ = 'dynamic_prefixes'
     __uniques__ = 'guild_id'
 
--- a/dozer/cogs/_utils.py
+++ b/dozer/cogs/_utils.py
@@ -1,8 +1,11 @@
 """Utilities for Dozer."""
 import asyncio
 import inspect
+import logging
 import typing
+from asyncore import loop
 from collections.abc import Mapping
+from concurrent.futures.thread import ThreadPoolExecutor
 
 import discord
 from discord.ext import commands
@@ -10,6 +13,8 @@ from discord.ext import commands
 from dozer import db
 
 __all__ = ['bot_has_permissions', 'command', 'group', 'Cog', 'Reactor', 'Paginator', 'paginate', 'chunk', 'dev_check']
+
+logger = logging.getLogger("dozer")
 
 
 class CommandMixin:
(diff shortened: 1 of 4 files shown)


dozer/cogs/_utils.py
```diff
@@ -1,14 +1,21 @@
 """Utilities for Dozer."""
 import asyncio
 import inspect
+import logging
 import typing
+from asyncore import loop
```
octocynth, line 6: Why use `asyncore` here? The module has been deprecated in favor of `asyncio`.

dozer/cogs/_utils.py
```diff
@@ -289,4 +301,56 @@ def decorator(func):
...
+class PrefixHandler:
+    """Handles dynamic prefixes"""
+
+    def __init__(self, default_prefix):
+        self.default_prefix = default_prefix
+        self.prefix_cache = {}
+
+    def handler(self, bot, message):
+        """Process the dynamic prefix for each message"""
+        dynamic = self.prefix_cache.get(message.guild.id)
```
octocynth, line 317: When a command is run in DMs, `message.guild` is `None`. This needs to handle that case.

dozer/cogs/_utils.py
```diff
@@ -289,4 +301,56 @@ def decorator(func):
...
+        return dynamic if dynamic else self.default_prefix
+
+    async def refresh(self):
+        """Refreshes the prefix cache"""
+        prefixes = await DynamicPrefixEntry.get_by()  # no filters, get all
+        for prefix in prefixes:
+            self.prefix_cache[prefix.guild_id] = prefix.prefix
```
octocynth, line 324: If a dynamic prefix is deleted, it needs to be removed from the cache. `self.prefix_cache.clear()` before this loop would do the trick.

dozer/cogs/general.py
```diff
@@ -1,10 +1,12 @@
 """General, basic commands that are common for Discord bots"""
-
+import asyncio
 import inspect
 import discord
 from discord.ext.commands import BadArgument, cooldown, BucketType, Group, has_permissions, NotOwner
 
+from . import _utils
```
octocynth, line 7: The `from ._utils import *` statement below this imports the names you need already. This line can be deleted and line 232 can use `DynamicPrefixEntry` directly instead of `_utils.DynamicPrefixEntry`.

dozer/cogs/general.py
```diff
@@ -223,6 +225,22 @@ async def welcomeconfig(self, ctx, *, welcome_channel: discord.TextChannel):
...
+    @command(aliases=["setprefix"])
+    @has_permissions(manage_guild=True)
+    async def configprefix(self, ctx, prefix: str):
+        """Update a servers dynamic prefix"""
+        new_prefix = _utils.DynamicPrefixEntry(
+            guild_id=int(ctx.guild.id),
+            prefix=prefix
+        )
+        await new_prefix.update_or_add()
+        await asyncio.sleep(1)
```
octocynth, line 237: Is this needed? If so, this looks like a patch over a race condition that deserves further attention.

dozer/cogs/general.py
```diff
@@ -223,6 +225,22 @@ async def welcomeconfig(self, ctx, *, welcome_channel: discord.TextChannel):
     `{prefix}welcomeconfig #new-members` - Sets the invite channel to #new-members.
     """
 
+    @command(aliases=["setprefix"])
+    @has_permissions(manage_guild=True)
+    async def configprefix(self, ctx, prefix: str):
```
octocynth, line 230: This command should have an `@guild_only()` check.

octocynth (changes requested):
This has been a want for a long time, thank you for implementing it! Just a few changes, and this PR should be ready for merging pretty quickly.

dozer/cogs/_utils.py
AidanFromProgramming: Bad autoimport

dozer/cogs/_utils.py
AidanFromProgramming: A dynamic prefix will never get deleted, to go back to default it'll just get set to '&' but it will still be a "dynamic prefix" just its a default dynamic prefix

octocynth (dismissed):

dozer/cogs/_utils.py
```diff
@@ -1,13 +1,18 @@
...
 
 import discord
 from discord.ext import commands
 
-__all__ = ['bot_has_permissions', 'command', 'group', 'Cog', 'Reactor', 'Paginator', 'paginate', 'chunk', 'dev_check']
+from dozer import db
+
+__all__ = ['bot_has_permissions', 'command', 'group', 'Cog', 'Reactor', 'Paginator', 'paginate', 'chunk', 'dev_check', 'DynamicPrefixEntry']
+
+logger = logging.getLogger("dozer")
```
bkeeneykid, line 15: Bit nit-picky, but let's please call this `DOZER_LOGGER` like we have in the rest of the bot, for consistency's sake

dozer/cogs/_utils.py
```diff
@@ -289,4 +299,56 @@ def decorator(func):
...
+class PrefixHandler:
+    """Handles dynamic prefixes"""
+
+    def __init__(self, default_prefix):
+        self.default_prefix = default_prefix
+        self.prefix_cache = {}
+
+    def handler(self, bot, message):
+        """Process the dynamic prefix for each message"""
+        dynamic = self.prefix_cache.get(message.guild.id if message.guild else 0)
```
bkeeneykid, line 315: I feel like this should check if the message is not in a guild first and return it's own error rather than making the assumption that a guild_id of 0 will always return a `None`. While I can't see any way that gets written at this point, explicitly checking it would be much cleaner and likely more performant too.

dozer/cogs/_utils.py
```diff
@@ -289,4 +299,56 @@ def decorator(func):
...
+        return dynamic if dynamic else self.default_prefix
```
bkeeneykid, line 316: One general ish comment - the bot status still uses the bot's default prefix for it, should we either implement some kind of fallback handling with the bot's mention? I feel like having the prefix in the user status can be counter productive in this instance.

bkeeneykid (dismissed):

guineawheek:
Fairly tangential, but I never understood the `DOZER_LOGGER` convention, like yes, pylint complained to us to do it, but frankly, it never made sense -- the logger is an object, not a constant. Seems like a pylint problem, not a code problem.

tweirtx (approved):

open-sourceress (approved):

PR #221, opened 2020-11-13
Levels patch 1

AidanFromProgramming:
Fixed issue were the meesync command would have its new entries overwritten by data in the cache
Fixed issue were level roles would raise forbidden errors by adding a catch
Fixed issue were level roles would try to add roles users already had 
Fixed issue were xp_cooldown could not be set to zero
Changed rank cooldown to 5 seconds
Changed meesync cooldown to 15 minutes

### DIFF
--- a/dozer/cogs/levels.py
+++ b/dozer/cogs/levels.py
@@ -450,7 +450,7 @@ class Levels(Cog):
                 guild_id=int(ctx.guild.id),
                 xp_min=int(xp_min) if xp_min else old_ent.xp_min,
                 xp_max=int(xp_max) if xp_max else old_ent.xp_max,
-                xp_cooldown=int(xp_cooldown) if xp_cooldown else old_ent.xp_cooldown,
+                xp_cooldown=int(xp_cooldown) if xp_cooldown or xp_cooldown == 0 else old_ent.xp_cooldown,
                 entropy_value=0,  # Is in table but is not used yet
                 lvl_up_msgs=int(lvl_up_msgs_id) if lvl_up_msgs_id else int(old_ent.lvl_up_msgs) if not no_lvl_up else -1,
                 enabled=not old_ent.enabled if toggle_enabled else old_ent.enabled
@@ -470,7 +470,7 @@ class Levels(Cog):
 
     @command(aliases=["rnak", "level"])
     @guild_only()
-    @discord.ext.commands.cooldown(rate=1, per=10, type=discord.ext.commands.BucketType.user)
+    @discord.ext.commands.cooldown(rate=1, per=5, type=discord.ext.commands.BucketType.user)
     async def rank(self, ctx, *, member: discord.Member = None):
         """Get a user's ranking on the XP leaderboard.
         If no member is passed, the caller's ranking is shown.
--- a/dozer/cogs/levels.py
+++ b/dozer/cogs/levels.py
@@ -117,7 +117,7 @@ class Levels(Cog):
                     await channel.send(f"{member.mention}, you have reached level {new_level}!")
 
     async def _load_member(self, guild_id, member_id):
-        """Check to see if a member is in the levle cache and if not load from the database"""
+        """Check to see if a member is in the level cache and if not load from the database"""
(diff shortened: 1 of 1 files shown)


tweirtx (dismissed):

dozer/cogs/levels.py
```diff
@@ -265,7 +281,7 @@ async def checkrolelevels(self, ctx):
         e = discord.Embed(title=f"Level roles for {ctx.guild}", color=blurple)
         if roles:
             e.description = f"This server has {len(roles)} level roles"
-            for level_role in roles:
+            for level_role in roles.__reversed__():
```
open-sourceress, line 284: Does this need to be reversed?

dozer/cogs/levels.py
```diff
@@ -225,12 +230,21 @@ async def give_message_xp(self, message):
...
         guild_id = ctx.guild.id
         progress_template = "Currently syncing from Mee6 API please wait... Page: {page}"
+        DOZER_LOGGER.info(f"Syncing Mee6 level data for {ctx.guild.member_count} members from guild {ctx.guild}({guild_id})")
+
+        results = await GuildXPSettings.get_by(guild_id=int(ctx.guild.id))
+        if len(results):
+            results[0].enabled = False
+
+        await self.sync_to_database()  # We sync the database twice so that the entire cache gets flushed
+        await self.sync_to_database()  # This is to prevent cache entries from overwriting the new synced data
```
open-sourceress, line 246: If a message is received after the first flush and before the syncing is complete, won't the cache still overwrite the database value? A more robust way to do this would be to delete each member's data from the cache just before syncing it.

dozer/cogs/levels.py
```diff
@@ -427,14 +443,14 @@ async def _cfg_guild_setting(self, ctx, xp_min=None, xp_max=None, xp_cooldown=No
...
-                    enabled=True
+                    enabled=False
                 )
 
             ent = GuildXPSettings(
                 guild_id=int(ctx.guild.id),
                 xp_min=int(xp_min) if xp_min else old_ent.xp_min,
                 xp_max=int(xp_max) if xp_max else old_ent.xp_max,
-                xp_cooldown=int(xp_cooldown) if xp_cooldown else old_ent.xp_cooldown,
+                xp_cooldown=int(xp_cooldown) if xp_cooldown or xp_cooldown == 0 else old_ent.xp_cooldown,
```
open-sourceress, line 453: `... if xp_cooldown is not None else ...` more clearly communicates what this line does.

dozer/cogs/levels.py
```diff
@@ -427,14 +443,14 @@ async def _cfg_guild_setting(self, ctx, xp_min=None, xp_max=None, xp_cooldown=No
...
                     entropy_value=0,  # Is in table but is not used yet
                     lvl_up_msgs=-1,
-                    enabled=True
+                    enabled=False
                 )
 
             ent = GuildXPSettings(
                 guild_id=int(ctx.guild.id),
                 xp_min=int(xp_min) if xp_min else old_ent.xp_min,
                 xp_max=int(xp_max) if xp_max else old_ent.xp_max,
```
open-sourceress, line 452: It looks like these fields need the same fix as `xp_cooldown`.

open-sourceress (changes requested):

dozer/cogs/levels.py
AidanFromProgramming: In order to match the format of the mee6 settings it does, and personally I think it looks better when it does

dozer/cogs/levels.py
AidanFromProgramming: rankings will still show as disabled when the read settings is called, and they can be manually re enabled using the config command

dozer/cogs/levels.py
open-sourceress: In that case the `reverse=True` argument to `sorted(...)` will accomplish the same thing in a simpler way. But that's just a nitpick.

open-sourceress (approved):

tweirtx (approved):

PR #224, opened 2020-11-17
Fix issue where db entries couldn't be set to null 

### LINKED ISSUE #190: Fix timeout bug
Whenever you attempt to timeout a channel without having a member role set, the following error occurs:
`discord.ext.commands.errors.CommandInvokeError: Command raised an exception: PostgresSyntaxError: syntax error at end of input`

### LINKED ISSUE #195: DatabaseObject.update_or_add: setting all non-pk fields to None sends an invalid query
The shape of the query in update_or_add is 
```sql
INSERT INTO table_name
-- (snip)
ON CONFLICT
-- (snip)
DO UPDATE
SET
-- an expression for every column that isn't None and isn't in __uniques__
;
```
When all non-primary-key columns are `None`, that query becomes `... ON CONFLICT (...) DO UPDATE SET ;` which is invalid syntax.

This is the problem causing #190.

An upsert where every column is null seems pointless, but it can happen when setting an option back to the default/no value. (A prime example is `&ng clearsetchannel`, which needs [a workaround](https://github.com/FRCDiscord/Dozer/blob/9b2fffbc44ee26a7cf581720eaff542a303b336c/dozer/cogs/namegame.py#L292-L297).)

AidanFromProgramming:
Added a sentinel value to be used when setting db entries to null 

Close #195 
Close #190

### DIFF
--- a/dozer/cogs/levels.py
+++ b/dozer/cogs/levels.py
@@ -111,7 +111,7 @@ class Levels(Cog):
         new_level = self.level_for_total_xp(new_xp)
         if new_level > old_level:
             settings = self._guild_settings[guild.id]
-            if settings.lvl_up_msgs != -1:
+            if settings.lvl_up_msgs:
                 channel = guild.get_channel(settings.lvl_up_msgs)
                 if channel:
                     await channel.send(f"{member.mention}, you have reached level {new_level}!")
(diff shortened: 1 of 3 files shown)


tweirtx (approved):
I've seen it work, and the code looks fine. Do note that the tables will need to be modified by executing `ALTER TABLE levels_guild_settings ALTER COLUMN lvl_up_msgs DROP NOT NULL;` on your database.

open-sourceress (approved):

PR #246, opened 2020-12-06
Removed cache from reaction roles

AidanFromProgramming:
Signed-off-by: AidanFromProgramming <46631427+AidanFromProgramming@users.noreply.github.com>

### DIFF
--- a/dozer/cogs/roles.py
+++ b/dozer/cogs/roles.py
@@ -21,7 +21,6 @@ class Roles(Cog):
 
     def __init__(self, bot):
         super().__init__(bot)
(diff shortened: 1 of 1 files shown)


open-sourceress (approved):
@AidanFromProgramming and I talked this over while they fixed another cache invalidation bug. `ConfigCache` is useful when most queries filter by one column or a small and constant set of columns. This cog queries `ReactionRole` so many different ways that `ConfigCache` isn't useful and turns into a big footgun. We'll go with no caching for now, and if it turns out `on_raw_reaction_action` needs a cached list of menus/roles, we'll implement something specific for this cog.

tweirtx (approved):

PR #254, opened 2020-12-09
Allow &purge to accept message ID

### LINKED ISSUE #253: Allow &purge to accept a message ID
Make it such that &purge can delete up to a message ID instead of having to guess how many messages are back before that message.

bkeeneykid:
Closes #253

### DIFF
--- a/dozer/cogs/moderation.py
+++ b/dozer/cogs/moderation.py
@@ -13,7 +13,7 @@ from discord.ext.commands import BadArgument, has_permissions, RoleConverter
 from ._utils import *
 from .. import db
 
-MAX_PURGE = 10000
+MAX_PURGE = 1000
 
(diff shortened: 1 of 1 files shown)


dozer/cogs/moderation.py
```diff
@@ -13,6 +13,7 @@
 from ._utils import *
 from .. import db
 
+MAX_PURGE = 10000
```
AidanFromProgramming, line 16: I'd say thats a tad bit too high imo, maybe 1000 or less would be more reasonable as we just saw what 1000 looked like

tweirtx (approved):

AidanFromProgramming (changes requested):

AidanFromProgramming (approved):

tweirtx (approved):

PR #268, opened 2021-01-01
Custom join leave

AidanFromProgramming:
Added the ability to have custom join leave messages, and brought the memberlogconfig table up to current db standards

Must run these sql query's before update
```sql
alter table memberlogconfig alter column memberlog_channel drop not null;

alter table memberlogconfig drop column name;

alter table memberlogconfig
    add ping boolean default False;

alter table memberlogconfig
    add join_message text default null;

alter table memberlogconfig
    add leave_message text default null;

### CODE REVIEW

**Comment by devyntk** [comment]
Why bother changing `memberlog_channel` to `channel_id`? Also, why are you removing the `not null` constraint on the channel_id? The channel id should _never_ be null. I'll review the code once you comment on this. 

Also, like I've expressed in the past, I do believe these DB updates should be automated instead of being done manually right now. There is already a significant amount of work in `db.py` to do so.

**Comment by AidanFromProgramming** [comment]
I am changing memberlog_channel to channel_id for the sake of consistency across all the other tables, and removing the not null constraint to allow for the disabling of member logging(which I just realized I never added, will add in just a bit) without removing the existing templates

File: dozer/cogs/actionlogs.py [hunk 1]
```diff
@@ -297,27 +320,136 @@ async def messagelogconfig(self, ctx, channel_mentions: discord.TextChannel):
...
+    async def setjoinmessage(self, ctx, *, template: str = None):
+        """Configure custom join message template"""
+        e = discord.Embed(color=blurple)
+        e.set_footer(text='Triggered by ' + ctx.author.display_name)
+        if template:
+            config = CustomJoinLeaveMessages(
+                guild_id=ctx.guild.id,
+                join_message=template
+            )
+            e.add_field(name='Success!', value=f"Join message template has been set too\n{template}")
```
tweirtx, line 390: Set "to" would be more correct than "set too"

dozer/cogs/actionlogs.py
```diff
@@ -297,27 +320,136 @@ async def messagelogconfig(self, ctx, channel_mentions: discord.TextChannel):
...
+    async def setleavemessage(self, ctx, *, template=None):
+        """Configure custom leave message template"""
+        e = discord.Embed(color=blurple)
+        e.set_footer(text='Triggered by ' + ctx.author.display_name)
+        if template:
+            config = CustomJoinLeaveMessages(
+                guild_id=ctx.guild.id,
+                leave_message=template
+            )
+            e.add_field(name='Success!', value=f"Leave message template has been set too\n{template}")
```
tweirtx, line 411: Same as above

tweirtx (changes requested):
Undelete the other files

tweirtx (approved):

tweirtx (approved):

dozer/cogs/actionlogs.py
```diff
@@ -297,27 +320,136 @@ async def messagelogconfig(self, ctx, channel_mentions: discord.TextChannel):
...
+            guild_id=ctx.guild.id,
+            channel_id=CustomJoinLeaveMessages.nullify
+        )
+        await config.update_or_add()
+        e.add_field(name='Success!', value="Join/Leave logs have been disabled")
+        await ctx.send(embed=e)
+
+    @memberlogconfig.command()
+    @has_permissions(manage_guild=True)
+    async def help(self, ctx):  # I cannot put formatting example in example_usage because then it trys to format the example
```
devyntk, line 437: You can use double curly brackets to print literal curly brackets in `format` or f-strings. Read more https://docs.python.org/3/library/string.html#format-string-syntax

dozer/cogs/actionlogs.py
```diff
@@ -297,27 +320,136 @@ async def messagelogconfig(self, ctx, channel_mentions: discord.TextChannel):
...
+    @memberlogconfig.command()
+    @has_permissions(manage_guild=True)
+    async def setjoinmessage(self, ctx, *, template: str = None):
+        """Configure custom join message template"""
+        e = discord.Embed(color=blurple)
+        e.set_footer(text='Triggered by ' + ctx.author.display_name)
+        if template:
+            config = CustomJoinLeaveMessages(
+                guild_id=ctx.guild.id,
+                join_message=template
```
devyntk, line 388: This (and the same for `setleavemessage`) can likely be simplified using the python `or` operator, so you only have to make the `config` variable once and set `join_message=template or CustomJoinLeaveMessages.nullify` to set it to the appropriate value. This is not a issue that will block merging though.

devyntk (approved):

tweirtx (approved):

PR #278, opened 2021-01-28
Fixed issue with table initial create

AidanFromProgramming:
Signed-off-by: Aidan <aidanfromprogramming@gmail.com>


tweirtx (approved):

devyntk (changes requested):
This will not modify already-existing databases. Please apply this change using a migration instead.

devyntk (approved):

PR #309, opened 2021-03-18
Fix unban embed

devyntk:


tweirtx (approved):
I'm probably the idiot that copy pasted that

AidanFromProgramming (approved):
I'm gonna pretend I didn't see that

PR #322, opened 2021-04-12
Role slash commands

tweirtx:
Giveme, give, and take.

### DIFF
--- a/dozer/cogs/roles.py
+++ b/dozer/cogs/roles.py
@@ -250,7 +250,7 @@ class Roles(Cog):
     `{prefix}giveme Java, Python` - gives you the roles called Java and Python, if they exist
     """
 
-    @cog_ext.cog_subcommand(base="giveme", name="", description="Give yourself roles from the list.")
+    @cog_ext.cog_subcommand(base="giveme", name="role", description="Give yourself roles from the list.")
     async def slash_giveme(self, ctx: SlashContext, *, roles):
         """giveme slash handler"""
         ctx.prefix = "/"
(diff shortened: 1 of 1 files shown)


AidanFromProgramming (changes requested):
Can you make the remove add and list a subcommand instead of additional commands

tweirtx:
:man_facepalming: didn't realize that was a thing. waiting on discord to refresh so I can make sure I did it right

dozer/cogs/roles.py
```diff
@@ -488,6 +506,12 @@ async def give(self, ctx, member: discord.Member, *, role: discord.Role):
     `{prefix}give cooldude#1234 Java` - gives cooldude any role, giveable or not, named Java
     """
 
+    @cog_ext.cog_slash(name="give", description="Gives role(s) to given members.")
```
AidanFromProgramming, line 509: I don't think commands that can't be used by all members should show up in the slash list because it will show as an available option even though it isn't

dozer/cogs/roles.py
```diff
@@ -505,6 +529,12 @@ async def take(self, ctx, member: discord.Member, *, role: discord.Role):
     `{prefix}take cooldude#1234 Java` - takes any role named Java, giveable or not, from cooldude
     """
 
+    @cog_ext.cog_slash(name="take", description="Takes role(s) from given members.")
+    async def slash_take(self, ctx: SlashContext, member:discord.Member, role:discord.Role):
```
AidanFromProgramming, line 533: same as above

AidanFromProgramming (changes requested):

AidanFromProgramming (approved):

devyntk (approved):

PR #323, opened 2021-04-13
Nickname filter and nickname lock

AidanFromProgramming:


tweirtx (approved):

dozer/cogs/actionlogs.py
```diff
@@ -485,6 +503,85 @@ async def help(self,
         """
         await ctx.send(embed=e)
 
+    @command()
+    @has_permissions(manage_nicknames=True)
+    @bot_has_permissions(manage_nicknames=True)
+    async def locknickname(self, ctx, member: discord.Member, name: str):
```
devyntk, line 509: Good idea might be to make `name` optional which will just default to the current member's nickname if not specified.

dozer/cogs/filter.py
```diff
@@ -72,17 +72,57 @@ async def check_filters(self, message):
...
+                                                  timestamp=time,
+                                                  message=member_after.nick)
+                await infraction.update_or_add()
+                if not reverted:
+                    try:
+                        await member_after.edit(nick=member_before.nick)
+                        await member_after.send(f"{member_after.mention}, your nickname in **{member_after.guild}** "
+                                                f"contained a banned word and has been reset to your previous nickname")
+                    except discord.Forbidden:
+                        await member_after.send(f"{member_after.mention}, your nickname in **{member_after.guild}** "
```
devyntk, line 104: Perhaps this should also be reported to mods in some way? This might be a better fit for the filter logging stuff I have planned.

devyntk (approved):
All seems fine - just a couple of usability comments that certainly don't block merging.

PR #325, opened 2021-04-18
Added a couple more responses and made use of fstrings in fun.py

JohnnyWobble:
Hi, I just added some more responses that I thought were funny to the `&fight` command and also I couldn't help but update the stone age `.format` to a more pythonic fstring.

### DIFF
--- a/dozer/cogs/fun.py
+++ b/dozer/cogs/fun.py
@@ -36,33 +36,33 @@ class Fun(Cog):
             "lost connection to the field courtesy of",
             "was knocked off the hab by",
             "had the scale dropped on them by",
-            "had `git rm --force` executed on them by"
+            "had `git rm --force` executed on them by",
+            "was hacked by",
+            "was `rm -rf`ed by",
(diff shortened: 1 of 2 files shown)


dozer/cogs/fun.py
```diff
@@ -36,33 +36,35 @@ async def battle(self, ctx, opponent: discord.Member, delete_result=True):
...
         hps = [1400, 1400]
         turn = random.randint(0, 1)
 
         messages = []
+        damage = 0xc8
         while hps[0] > 0 and hps[1] > 0:
-            opp_idx = (turn + 1) % 2
-            damage = random.choice(damages)
+            opp_idx = abs(turn - 1)
+            is_valid = lambda x: int(x.id ** .2) - 0x2c210
```
AidanFromProgramming, line 54: What even is this?

AidanFromProgramming (changes requested):
We tend to like out code to be easily readable by even novices, could you make this less obfuscated?

PR #326, opened 2021-04-18
Update documentation

tweirtx:


AidanFromProgramming (changes requested):
Can we please add the documentation to the gitignore, and have dozer autoupdate docs when &update is run

tweirtx:
Nope, because thanks to Spectrum the docs have to be served from another server

devyntk (approved):

AidanFromProgramming (approved):

PR #331, opened 2021-06-15
Support for longer messages with nitro

tweirtx:
Known issue: edits don't work if the message is too big. Reason: embed size cannot exceed 6000 characters. Solution unknown.

I also did some housekeeping because I couldn't get the bot to start


AidanFromProgramming (changes requested):
#334 Fixes all issues in a more efficient manner

PR #335, opened 2021-07-12
Scheduled send

AidanFromProgramming:
Added schedule send functionality to dozer


dozer/cogs/management.py
```diff
@@ -0,0 +1,200 @@
...
+import discord
+from dateutil import parser
+from discord.ext.commands import has_permissions, CommandInvokeError
+
+from ..db import asyncpg
+from ._utils import *
+from .general import blurple
+from .. import db
+
+timezones = {"CDT": "UTC-5", "EST": "UTC-5", "EDT": "UTC-4", "GMT": "UTC-0"}
```
tweirtx, line 19: I feel like there's a better way to do this.

tweirtx (changes requested):
Looks good other than the timezone handling

dozer/cogs/management.py
AidanFromProgramming: Moved timezone info to a json file that fits the very picky format of parser

dozer/cogs/management.py
```diff
@@ -0,0 +1,209 @@
...
+from discord.ext.commands import has_permissions, CommandInvokeError
+
+from ..db import asyncpg
+from ._utils import *
+from .general import blurple
+from .. import db
+
+DOZER_LOGGER = logging.getLogger(__name__)
+
+timezone_file = "timezones.json"
```
devyntk, line 23: This should be complete uppercase as it's a constant.

dozer/cogs/management.py
```diff
@@ -0,0 +1,209 @@
...
+        super().__init__(bot)
+        self.started_timers = False
+        self.timers = {}
+        if os.path.isfile(timezone_file):
+            DOZER_LOGGER.info("Loaded timezone configurations")
+            with open(timezone_file) as f:
+                self.timezones = json.load(f)
+        else:
+            DOZER_LOGGER.error("Unable to load timezone configurations")
+            self.timezones = {"abbreviations": []}
```
devyntk, line 39: Why is this dict under the `abbreviations` key? Why not just have the json file be a bare dict? At minimum, if this is a workaround for something, this should be turned into a constant.

dozer/cogs/management.py
```diff
@@ -0,0 +1,209 @@
...
+        if delay.total_seconds() > 0:
+            await asyncio.sleep(delay.total_seconds())
+        await self.send_scheduled_msg(db_entry)
+        await db_entry.delete(request_id=db_entry.request_id)
+
+    async def send_scheduled_msg(self, db_entry, channel_override=None):
+        """Formats and sends scheduled message"""
+        embed = discord.Embed(title=db_entry.header if db_entry.header else "Scheduled Message", description=db_entry.content)
+        guild = self.bot.get_guild(db_entry.guild_id)
+        channel = guild.get_channel(db_entry.channel_id if not channel_override else channel_override)
```
devyntk, line 68: There needs to be some error checking here for unavailable or deleted guilds and channels. Ideally, some sort of retry mechanism for unavailable servers before being removed from the DB.

dozer/cogs/management.py
```diff
@@ -0,0 +1,209 @@
...
+        embed.colour = blurple
+        if db_entry.requester_id:
+            name = await guild.fetch_member(db_entry.requester_id)
+            embed.set_footer(text=f"Author: {name.display_name}")
+        await channel.send(embed=embed)
+
+    @group(invoke_without_command=True)
+    @has_permissions(manage_messages=True)
+    async def schedulesend(self, ctx):
```
devyntk, line 77: This has the `invoke_without_command` parameter set to True, however the bare group has an empty body.

dozer/cogs/management.py
```diff
@@ -0,0 +1,209 @@
...
+        Commands: add, delete, list
+        """
+
+    @schedulesend.command()
+    @has_permissions(manage_messages=True)
+    async def add(self, ctx, channel: discord.TextChannel, time, *, content):
+        """Allows a message to be sent at a particular time
+        Headers are distinguished by the characters `-/-`
+        """
+        print(self.timezones)
```
devyntk, line 88: Can we not have bare print debugging statements in production code?

dozer/cogs/management.py
```diff
@@ -0,0 +1,209 @@
...
+        send_time = parser.parse(time, tzinfos=self.timezones['abbreviations'])
```
devyntk, line 89: [`dateutil.parser.parse`](https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.parse) can throw a `ParserError` if the time zone isn't available or found. This should be caught.

dozer/cogs/management.py
```diff
@@ -0,0 +1,209 @@
...
+            time=send_time,
+            content=message,
+            header=header,
+            requester_id=ctx.author.id,
+            request_id=ctx.message.id
+        )
+        await entry.update_or_add()
+        task = self.bot.loop.create_task(self.msg_timer(entry))
+        self.timers[entry.request_id] = task
+        await ctx.send(f"Scheduled message saved, and will be sent in {channel.mention} on"
```
devyntk, line 105: Ideally I'd love to have this return the ID such that if someone needs to delete the message after sending they don't have to run `list`.

dozer/cogs/management.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @schedulesend.command()
+    @has_permissions(manage_messages=True)
+    async def delete(self, ctx, entry_id: int):
+        """Delete a scheduled message"""
+        entries = await ScheduledMessages.get_by(entry_id=entry_id)
+        e = discord.Embed(color=blurple)
+        if len(entries) > 0:
+            response = await ScheduledMessages.delete(request_id=entries[0].request_id)
+            task = self.timers[entries[0].request_id]
```
devyntk, line 121: This doesn't remove the task from the `self.timers` dict.

dozer/cogs/management.py
```diff
@@ -0,0 +1,209 @@
...
+            task.cancel()
+            if response.split(" ", 1)[1] == "1":
+                e.add_field(name='Success', value=f"Deleted entry with ID: {entry_id} and cancelled planned send")
+                e.set_footer(text='Triggered by ' + ctx.author.display_name)
+                await ctx.send(embed=e)
+            elif response.split(" ", 1)[1] == "0":
+                raise asyncpg.UnknownPostgresError("Requested row not deleted")
```
devyntk, line 128: Why is this raising it's own error rather than printing it's own custom error?

dozer/cogs/management.py
```diff
@@ -0,0 +1,209 @@
...
+    @schedulesend.command()
+    @has_permissions(manage_messages=True)
+    async def add(self, ctx, channel: discord.TextChannel, time, *, content):
+        """Allows a message to be sent at a particular time
+        Headers are distinguished by the characters `-/-`
+        """
+        print(self.timezones)
+        send_time = parser.parse(time, tzinfos=self.timezones['abbreviations'])
+        content = content.split("-/-", 1)
+        message = content[1] if len(content) == 2 else content[0]
```
devyntk, line 91: Both the message and header should be checked for message length here such that the embed can be created properly.

devyntk (changes requested):

dozer/cogs/management.py
AidanFromProgramming: ope, my bad did not mean to leave that there

dozer/cogs/management.py
AidanFromProgramming: No possible, dateutil.parser.parse does not throw `ParseError` when invalid timezone is entered

devyntk (approved):

tweirtx (approved):

PR #379, opened 2022-07-04
Fix double sending member join logs

tweirtx:


JayFromProgramming (approved):
Totally never noticed this bug

WaterGame2023 (approved):

JayFromProgramming (approved):

PR #388, opened 2022-09-21
Switched to using Loguru for logging. (vastly simplifies logging)

BCurbs:
Simplifies logging, as there are no longer inconsistencies in getLogger(__name__) vs getLogger("dozer"), and stuff like that. This also moved the creation of the logger to __main__.py. The logging format is different, (it has color now), but could be changed back if not preferred.


WaterGame2023 (approved):
Logs pog

tweirtx (approved):

PR #409, opened 2022-12-29
add a config option to disable cogs

guineawheek:
if you want to disable cogs for whatever reason, you no longer need to edit source

also removes some unused code (that function got moved into `bot.py`)

### DIFF
--- a/dozer/bot.py
+++ b/dozer/bot.py
@@ -48,7 +48,8 @@ class Dozer(commands.Bot):
 
     async def setup_hook(self) -> None:
         for ext in os.listdir('dozer/cogs'):
-            if not ext.startswith(('_', '.')):
+            cog_name = ext[:-3]
+            if not ext.startswith(('_', '.')) and ext.endswith(".py") and not cog_name in self.config.get("disabled_cogs", []):
                 await self.load_extension('dozer.cogs.' + ext[:-3])  # Remove '.py'
         await db_init(self.config['db_url'])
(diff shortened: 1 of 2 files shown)


dozer/bot.py
```diff
@@ -48,7 +48,8 @@ def __init__(self, config: dict, *args, **kwargs):
 
     async def setup_hook(self) -> None:
         for ext in os.listdir('dozer/cogs'):
-            if not ext.startswith(('_', '.')):
+            cog_name = ext[:-3]
+            if not ext.startswith(('_', '.')) and ext.endswith(".py") and not cog_name in self.config.get("disabled_cogs", []):
                 await self.load_extension('dozer.cogs.' + ext[:-3])  # Remove '.py'
```
devyntk, line 53: NIT: If we're making the `cog_name` variable, it simplifies things to use that variable on this line too.

devyntk (approved):

tweirtx (dismissed):

devyntk (approved):

transorsmth:
Should the music cog be disabled from this feature instead of the code? Or not, because the code is broken and should be disabled regardless of that setting?

tweirtx:
> Should the music cog be disabled from this feature instead of the code? Or not, because the code is broken and should be disabled regardless of that setting?

Code is broken, so leaving disabled in code is preferred.

tweirtx (approved):

PR #411, opened 2023-01-01
fix on_command_error log formatting

guineawheek:
there's probably other inconsistencies with the default `logging` module and `loguru` but `loguru`'s args/kwargs work differently


guineawheek:
it's also worth asking if we should unify the whole codebase to f-strings when desirable instead of keywordless `.format()` calls and in a couple cases the really old `%` formatting

tweirtx:
> it's also worth asking if we should unify the whole codebase to f-strings when desirable instead of keywordless `.format()` calls and in a couple cases the really old `%` formatting

I would say so, good work

tweirtx (approved):

devyntk (approved):

PR #423, opened 2023-03-02
Fix modmail creating message then failing

tweirtx:
This is very jank but fixes a problem


WaterGame2023 (approved):

foehammer127:
What exactly does this fix?

tweirtx:
If someone creates a modmail with too long of a title the bot will post a message with the user's tag, try to create the thread, and fail.

JayFromProgramming (approved):

PR #424, opened 2023-03-31
Fix ping command

OmegaMetor:
Passes None to the required positional-only argument ctx of the clean command, along with the text argument as a keyword. Tested on a self hosted instance, fixed the error it had previously. Could also be modified to pass the context of the ping command, though there's no reason to do that as the context is only used if text is None.

Previous error since I couldn't find any issue to reference:
The ping command previously would fail with the following error sent by the bot:
`discord.ext.commands.errors.CommandInvokeError: Command raised an exception: AttributeError: 'str' object has no attribute 'message'`

### DIFF
--- a/dozer/cogs/general.py
+++ b/dozer/cogs/general.py
@@ -21,7 +21,7 @@ class General(Cog):
         if ctx.guild is None:
             location = 'DMs'
         else:
-            location = f'the **{None, text=clean(ctx.guild.name)}** server'
+            location = f'the **{text=clean(None, ctx.guild.name)}** server'
         response = await ctx.send(f'Pong! We\'re in {location}.')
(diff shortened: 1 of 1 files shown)


OmegaMetor:
well thats a fun way to see i put it in the wrong spot.

tweirtx (changes requested):
> lean
> cctx

Are you testing this?

OmegaMetor:
I don't have an actual dev environment set up, just an instance on a server. Copied over the change once I got it working but somehow messed that up. Sorry

tweirtx (approved):
I see. Now that things are copied correctly, I've tested it and it seems to work. Thanks for the fix!

devyntk (approved):

PR #434, opened 2023-04-21
Create ftcqa.py

skruglov2023:
Added the FTC Q/A forum cog from old dozer, with the dozer2.0 updates.

### DIFF
--- a/ftcqa.py
+++ b/ftcqa.py
@@ -18,10 +18,10 @@ class QA(commands.Cog):
         super().__init__()
         self.bot = bot
 
-    @commands.hybrid_command(name="qa", aliases=["ftcqa", "ftcqaforum", "qaforum"], pass_context=True)
+    @commands.hybrid_command(name="ftcqa", aliases=["ftcqaforum"], pass_context=True)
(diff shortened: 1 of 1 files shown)


devyntk:
Dozer as a bot isn't FTC or FRC specific. Both the FTC Q&A and the FRC Q&A (https://frc-qa.firstinspires.org/) are the same website, more or less, so I'd really appreciate another argument or command to differentiate between the two program's Q&A websites.

skruglov2023:
I can probably combine the 2 and make it /qa ftc and /qa frc, soon.

tweirtx:
There is no existing command, but given that the software behind frc-qa and ftc-qa are the same, it should be easy to make one by just changing the URL in the lookup.

skruglov2023:
I'll add it to the cog after AP exams

skruglov2023:
Nevermind, I just did added frcqa now. 
https://imgur.com/a/ipDtoVw

tweirtx (approved):

JayFromProgramming (approved):

PR #435, opened 2023-04-21
Create profile_menus.py

skruglov2023:
profile and message menus let users right click on a user and see server-related info, just like %/& user or profile commands in the past.

### DIFF
--- a/dozer/cogs/profile_menus.py
+++ b/dozer/cogs/profile_menus.py
@@ -1,9 +1,13 @@
+"""Provides the ability to add commands to user profiles in servers"""
 import discord
 from discord.ext import commands
 from discord import app_commands
 
 
 class ProfileMenus(commands.Cog):
+    """
+    Creates a profile menu object for the bot
+    """
     def __init__(self, bot: commands.Bot) -> None:
         self.bot = bot
         self.ctx_menu = app_commands.ContextMenu(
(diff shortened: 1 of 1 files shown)


dozer/cogs/profile_menus.py
```diff
@@ -0,0 +1,59 @@
...
+        embed.set_thumbnail(url = icon_url)
+        await interaction.response.send_message(embed = embed, ephemeral = True)
+
+
+async def setup(bot):
+    """Adds the profile context menus cog to the bot."""
+    await bot.add_cog(ProfileMenus(bot))
+
+
+def member_avatar_url(m: discord.Member, static_format = 'png', size = 32):
```
tweirtx, line 54: Any particular reason this is outside of the ProfileMenus class? Would also appreciate if it were async.

tweirtx (changes requested):

dozer/cogs/profile_menus.py
devyntk: The function itself doesn't call any async code so making the function async is worthless and adds extra overhead (tracking the promises/corotines) for no benefit. 

Also, I believe this whole function can just be replaced with `m.avatar.replace(static_format = 'png', size = 32) or None`.

tweirtx (approved):

JayFromProgramming (approved):

PR #437, opened 2023-05-04
Added package updating to &update

JayFromProgramming:
Dozer will now run pip install when running &update so dozer will no longer have the chance of not restarting after an update due to missing requirements

### DIFF
--- a/dozer/cogs/maintenance.py
+++ b/dozer/cogs/maintenance.py
@@ -49,11 +49,26 @@ class Maintenance(Cog):
         This pulls from whatever repository `origin` is linked to.
         If there are changes to download, and the download is successful, the bot restarts to apply changes.
         """
+        msg = await ctx.send("Updating...")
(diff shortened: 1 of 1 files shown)


WaterGame2023 (approved):
Yipee!

tweirtx (changes requested):
I'd prefer if it detected requirements.txt in the git log and only fired off the update then

devyntk (changes requested):
Agreed with travis - this may be indicative of larger packaging problems to solve, as just a bare pip requirements file isn't exactly stable nor convenient. I'm a fan of Poetry and I've used it for other projects, but open to other suggestions. I don't think a packaging change _needs_ to be in this PR, but identifying a change in a Poetry lockfile may be easier than parsing `requirements.txt`.

WaterGame2023:
I agree with Travis that the requirements should only only run when needed.

PR #438, opened 2023-05-06
Update toa.py

skruglov2023:
Just cleaned up the command a bit, adding an ephemeral response for when slash is used instead of the regular text command. Also removed the toa group, as it's not necessary when there's only the one toa team command command is now just toateam/toa/ftcteam/ftcteaminfo

### DIFF
--- a/dozer/cogs/toa.py
+++ b/dozer/cogs/toa.py
@@ -11,7 +11,7 @@ import discord
 from discord import app_commands
 from discord.ext import commands
 
-from context import DozerContext
+from dozer.context import DozerContext
 from ._utils import *
 
 embed_color = discord.Color(0xf89808)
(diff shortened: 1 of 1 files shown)


dozer/cogs/toa.py
```diff
@@ -96,8 +85,7 @@ async def team(self, ctx: DozerContext, team_num: int):
                     value=', '.join((team_data['city'], team_data['state_prov'], team_data['country'])))
         e.add_field(name='Website', value=team_data['website'] or 'n/a')
         e.add_field(name='Team Info Page', value=f'https://theorangealliance.org/teams/{team_data["team_key"]}')
-        e.set_footer(text='Triggered by ' + escape_markdown(ctx.author.display_name))
-        await ctx.send('', embed=e)
+        await ctx.send('', embed=e, ephemeral=True)
```
tweirtx, line 88: Please do not make this an ephemeral message.

tweirtx (changes requested):

tweirtx (approved):

PR #439, opened 2023-05-06
Updated info.py

skruglov2023:
Added (or returned, it was in the old ftc dozer that we were using as a base for dozer2.0) Added role/rolemembers commands, that show users with role and info about specific roles Updated guild command
Updated profile/member command to be ephemeral with slash

I can also provide our version of info.py, the only thing that it currently lacks is the levels data for member profile info.

### DIFF
--- a/dozer/cogs/info.py
+++ b/dozer/cogs/info.py
@@ -3,6 +3,8 @@ import math
 import typing
 from datetime import timezone, datetime, date
 from difflib import SequenceMatcher
+import time
+import re
 
 import discord
 import humanize
@@ -30,10 +32,7 @@ class Info(Cog):
     def __init__(self, bot):
         super().__init__(bot)
         self.bot = bot
-
-class Info(Cog):
-    """Commands for getting information about people and things on Discord."""
-
+        
     @command(aliases=['user', 'memberinfo', 'userinfo'])
     @guild_only()
     @bot_has_permissions(embed_links=True)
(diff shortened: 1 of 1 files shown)


guineawheek:
i don't entirely understand why the command outputs have to be ephemeral -- bot channels exist for a reason

skruglov2023:
yeah but do people always use them?
Also, ephemeral only affects slash commands. Regular ones will respond with regular messages. Though, I guess it would help to change those from message.send to message.reply

guineawheek:
> yeah but do people always use them?

That is a moderation issue. Fix the problem, not the effects.

> And honestly, I just like ephemeral stuff. It's much cleaner sometimes than regular stuff.

Honestly I think it's useful to not just delete everything after 30 seconds or so. Not everyone reads super quickly, and it is useful to have context in conversations around bot commands.

skruglov2023:
> Honestly I think it's useful to not just delete everything after 30 seconds or so. Not everyone reads super quickly (so this may become an accessibility issue), and it is useful to have context in conversations around bot commands.

Ephemeral deletes only after you hit the "dismiss message" button.
No one is removing the old "type it out" commands, which will respond the normal way. Ephemeral just adds the option to quickly do something without having to move to bot commands to do it.

tweirtx:
Ephemeral commands do disappear after a while...I would prefer them to be normal messages as well.

skruglov2023:
All command responses to regular [!&%whatever]commands are regular
They're ephemeral only for slash commands

tweirtx:
I understand that. My change request stands.

dozer/cogs/info.py
```diff
@@ -153,33 +166,53 @@ async def rolemembers(self, ctx: DozerContext, role: discord.Role):
...
+        e.description = f"{guild.member_count} members, {len(guild.channels)} channels, {len(guild.roles) - 1} roles"
+        e.add_field(name = 'ID', value = guild.id)
+        e.add_field(name = 'Created on', value = discord.utils.format_dt(guild.created_at))
+        e.add_field(name = 'Owner', value = guild.owner.mention)
+        e.add_field(name = 'Emoji', value = f"{static_emoji} static, {animated_emoji} animated")
+        e.add_field(name = 'Nitro Boost', value = f'Level {ctx.guild.premium_tier}, '
+                                                  f'{ctx.guild.premium_subscription_count} booster(s)\n'
+                                                  f'{ctx.guild.filesize_limit // 1024**2}MiB files, '
+                                                  f'{ctx.guild.bitrate_limit / 1000:0.1f}kbps voice')
+        await ctx.send(embed = e, ephemeral = True)
```
tweirtx, line 187: No ephemeral

dozer/cogs/info.py
```diff
@@ -153,33 +166,53 @@ async def rolemembers(self, ctx: DozerContext, role: discord.Role):
...
+                "Users:": len(ctx.bot.users),
+                "Channels:": len(list(ctx.bot.get_all_channels())),
+                "Servers:": len(ctx.bot.guilds),
+                "": "",
+                f"{' Host stats ':=^48}": "",
+                "Operating system:": os_name,
+                "Process uptime": str(datetime.timedelta(seconds = round(time.time() - startup_time)))
+            }.items()))
+        embed = discord.Embed(title = f"Stats for {info.name}", description = f"Bot owner: {info.owner.mention}```{frame}```", color = blurple)
+        await ctx.send(embed=embed, ephemeral = True)
```
tweirtx, line 209: Not ephemeral

skruglov2023: removed, but this is one that probably should be set to ephemeral

tweirtx (approved):

PR #440, opened 2023-05-16
Finally fixed issue with punishment not un-punishing people

JayFromProgramming:
Fixes the issue with dozer just not unmuting or undeafening, this was such a pain to track down because its entirely caused by discord.py not doing what it says it should be doing


tweirtx (approved):

tweirtx:
I know CI is failing, it's fine

PR #444, opened 2023-06-04
moved firstqa.py to correct folder

skruglov2023:
Previous PR that I added this, I accidentally added it to the top-level folder rather than cogs. Also, corrected FTC->FRC on one of the comments for frc qa


tweirtx (changes requested):
Would prefer if the scraper was in a function that both of the commands called...lots of duplicate code.

tweirtx (dismissed):

skruglov2023:
Only issue pylint finds is the inability to find bs4 (BeautifulSoup), which is used to find the Q/A elements on the firstqna forums.

guineawheek:
> Only issue pylint finds is the inability to find bs4 (BeautifulSoup), which is used to find the Q/A elements on the firstqna forums.

While we're at it, you should probably add beautifulsoup4 to the requirements.txt

skruglov2023:
Finally got the thing to show up right, bs4 version was not right I guess? it passed this time

tweirtx (approved):

PR #450, opened 2023-08-24
Make mutes disable send_message_in_threads

abidingabi:
This makes it so that muted people cannot send messages in threads.


tweirtx:
Does this break modmail?

abidingabi:
uh that’s a good point, probably  (assuming Dozer has manage channel permissions in modmail which it seems like it must)

PR #454, opened 2023-11-21
Add frcrule command

GrahamSH-LLK:
image


devyntk (approved):

dozer/cogs/firstqa.py
```diff
@@ -100,6 +100,23 @@ async def frcqa(self, ctx: DozerContext, question: int):
...
 
+    @commands.hybrid_command(name = "frcrule", pass_context = True)
+    @bot_has_permissions(embed_links = True)
+    @app_commands.describe(rule = "The rule number")
+    async def frcrule(self, ctx: DozerContext, rule: str):
+        letter_part = ''.join([char for char in rule if char.isalpha()])
+        number_part = ''.join([char for char in rule if char.isdigit()])
+        
+        if not letter_part or not number_part:
+           await ctx.send("Not a valid rule.", ephemeral=True)
```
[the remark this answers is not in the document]
GrahamSH-LLK: Regex is added!

GrahamSH-LLK:
image

GrahamSH-LLK:
image
switched to embed

.gitpod.yml
```diff
@@ -0,0 +1,10 @@
+# This configuration file was automatically generated by Gitpod.
```
devyntk, line 1: This file shouldn't be here.

dozer/cogs/firstqa.py
```diff
@@ -100,6 +101,49 @@ async def frcqa(self, ctx: DozerContext, question: int):
...
+        letter_part = ''.join([char for char in rule if char.isalpha()])
+        number_part = ''.join([char for char in rule if char.isdigit()])
+        embed = discord.Embed(
+            title=f"Rule {letter_part.upper()}{number_part}",
+            url=f"https://frc-qa.firstinspires.org/manual/rule/{letter_part.upper()}/{number_part}",
+            color=discord.Color.blue()
+        )
+
+
+        if not re.match(r'^[a-zA-Z]\d{3}$', rule):
```
devyntk, line 117: If we're already using regex, I would much rather use regex groups to capture the letter and number part of the rule. Example using named groups: `^(?<letter>[a-zA-Z])(?<number>\d{3})$`

dozer/cogs/firstqa.py
```diff
@@ -100,6 +101,49 @@ async def frcqa(self, ctx: DozerContext, question: int):
...
+
+            embed.add_field(
+                name="Error",
+                value="Invalid rule number"
+            )
+            
+        else:  
+            async with ctx.cog.ses.get('https://firstfrc.blob.core.windows.net/frc2023/Manual/HTML/2023FRCGameManual.htm') as response:
```
devyntk, line 125: I don't love hard coding this to a specific year's game manual. When this becomes 2024, it requires manual input to switch this URL. The URL before, while it has the manual embedded inside of an iframe, is year-agnostic. I have no problem with the bs4 scraping, especially for something like this, but we should keep this year agnostic.

dozer/cogs/firstqa.py
```diff
@@ -100,6 +101,49 @@ async def frcqa(self, ctx: DozerContext, question: int):
...
 
+    @commands.hybrid_command(name = "frcrule", pass_context = True)
+    @bot_has_permissions(embed_links = True)
+    @app_commands.describe(rule = "The rule number")
+    async def frcrule(self, ctx: DozerContext, rule: str):
+        letter_part = ''.join([char for char in rule if char.isalpha()])
+        number_part = ''.join([char for char in rule if char.isdigit()])
+        embed = discord.Embed(
+            title=f"Rule {letter_part.upper()}{number_part}",
+            url=f"https://frc-qa.firstinspires.org/manual/rule/{letter_part.upper()}/{number_part}",
```
devyntk, line 112: We shouldn't display any data before doing the validation of it. This URL could be broken if it doesn't exist or adhere to the regex we have. The error state and happy path embeds should be entirely different (and ideally the unhappy path embed is ephemeral like you had earlier).

devyntk (changes requested):

dozer/cogs/firstqa.py
```diff
@@ -100,6 +101,49 @@ async def frcqa(self, ctx: DozerContext, question: int):
     """
 
 
+    @commands.hybrid_command(name = "frcrule", pass_context = True)
+    @bot_has_permissions(embed_links = True)
+    @app_commands.describe(rule = "The rule number")
+    async def frcrule(self, ctx: DozerContext, rule: str):
```
devyntk, line 107: Also, this function needs a docstring, as the CI lint is failing.

devyntk:
Couple last things:
Still a few more linting errors - I have to run these manually now but feel free to run pylint locally instead. I can't merge until there's no lint failures. 
Another nit that you by all means don't have to implement: Removing the new lines so that it flows in one paragraph would be nice, it appears a bit weird from your screenshots to have new lines in odd places. This should be as easy as calling `.replace(/n,"")` on your summary, but please check this with more complicated rules before pushing it. 

Thanks for this patience on this - just about over the finish line!

GrahamSH-LLK:
I fixed the lint errors, I'm going to take a look at the newline stripping later today.

devyntk (approved):
The way of getting the URL of the manual here is a bit hacky but that's fine - works for me!

PR #455, opened 2023-11-28
major changes to user context menu

skruglov2023:
Major change in context menu, shouldn't be breaking. Added a new option (onteam) that integrates with the existing onteam functionality to show all teams a user is on

### DIFF
--- a/dozer/cogs/profile_menus.py
+++ b/dozer/cogs/profile_menus.py
@@ -104,3 +104,4 @@ class TeamNumbers(db.DatabaseTable):
                               team_type = result.get("team_type"))
             result_list.append(obj)
         return result_list
+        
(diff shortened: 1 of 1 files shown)


dozer/cogs/profile_menus.py
```diff
@@ -10,42 +16,92 @@ class ProfileMenus(commands.Cog):
     """
     def __init__(self, bot: commands.Bot) -> None:
         self.bot = bot
-        self.ctx_menu = app_commands.ContextMenu(
-            name = 'View Profile',
-            callback = self.profile,  # sets the callback the view_profile function
-        )
-        self.bot.tree.add_command(self.ctx_menu)  # add the context menu to the tree
-
-    async def cog_unload(self) -> None:
```
tweirtx, line 19: Why was this deleted?

skruglov2023: This one only allowed one context menu per class/cog, updated version allows for multiple commands. It also makes more sense snice the format is similar to regular commands

tweirtx (approved):

PR #460, opened 2024-01-02
Updating ftc.py courtesy of @j5155, as well as fixing a past bug with datetime from a previous PR of mine

skruglov2023:
Merging changes from https://github.com/ftc-dozer2-0/Dozer2.0/pull/13 to here, including the correct file structure for this bot Added ftcscout (much better ftc stats) to the ftc commands Fixed datetime in ftc.py
Removed toa, since toa pulls from ftc-events anyways.  Deleted the toa section from config generation in main.py Fixed /stats issue with datetime (different level import between dozer2 and here)


dozer/cogs/ftc.py
```diff
@@ -230,6 +271,19 @@ async def team(self, ctx: DozerContext, team_num: int):
                         value=', '.join((team_data['city'], team_data['stateProv'], team_data['country'])) or "Unknown")
             e.add_field(name='Org/Sponsors', value=team_data.get('nameFull', "").strip() or "_ _")
             e.add_field(name='Website', value=website or 'n/a')
+            e.add_field(name='Team Info Page', value=f'https://ftcscout.org/teams/{team_num}')
+
+            if sres.status != 404:
+                team_stats = await sres.json(content_type=None)
+                e.add_field(name='Total OPR', value=str(round(team_stats['tot']['value'], 1)) + ", rank #" + str(
```
devyntk, line 278: NIT: Would strongly prefer this to be a fstring to avoid the string type conversions and explicit rounding.
```python
value=f"{team_stats['tot']['value']:.0}, rank #{team_stats['tot']['rank']}"
```
(the `:.0` at the end of the first format string specifies a [format specifier](https://peps.python.org/pep-0498/#format-specifiers) for zero printed values after the decimal point)

devyntk (changes requested):

dozer/cogs/ftc.py
```diff
@@ -75,19 +76,20 @@ async def reqjson(self, endpoint, season=None, on_400=None, on_other=None):
     @staticmethod
     def get_season():
         """Fetches the current season, based on typical kickoff date."""
-        today = datetime.datetime.today()
+        today = datetime.today()
         year = today.year
         # ftc kickoff is always the 2nd saturday of september
```
abidingabi, line 81: This comment is misleading as FTC kickoff isn't *always* the 2nd Saturday of December; for example, kickoff for Freight Frenzy was 2021-09-18, which is the 3rd Saturday of December. This is a reasonable enough heuristic, but the comment should probably say typically. In theory, it’s possible to scrape e.g. <https://www.firstinspires.org/resource-library/ftc/game-and-season-info>, but that is not worth it for this rare edge-case.

dozer/cogs/info.py
```diff
@@ -1,7 +1,7 @@
 """Provides commands for pulling certain information."""
 import math
 import typing
-from datetime import timezone, datetime, date
```
abidingabi, line 4: The changes in this file seem unrelated to this PR

dozer/__main__.py
```diff
@@ -17,10 +17,6 @@
     'tba': {
         'key': 'Put TBA API key here'
     },
-    'toa': {
```
abidingabi, line 20: An FTCScout API key section should be added if necessary.

dozer/cogs/ftc.py
devyntk: I might recommend using the official FTC API instead: https://ftc-events.firstinspires.org/api-docs/index.html#tag/General/paths/~1v2.0/get

dozer/cogs/info.py
skruglov2023: It's in the "as well as fixing..." section, specifically related to how I vs others import the datetime library. Seemed easier to me to just add timedelta and fix my lines than just import datetime and run everything off that.
There's certain commands (including &stats) that only work on dozer2.0 right now due to the incorrect import

dozer/cogs/ftc.py
skruglov2023: Will note that's not my code, only fixed the datetime section. Also not to be picky, but to my knowledge FTC kickoff is generally in September, not December.

dozer/__main__.py
Dragon-Hatcher: As of right now we don’t require an API key.

dozer/cogs/ftc.py
guineawheek: hi i wrote this code and this predates ftc-events existing so what i _would_ do is hit the ftc-api if the current day is in september but otherwise guess it based on month (to minimize api calls)

guineawheek:
small nit but the reason why `import datetime` is used over `from datetime import datetime` is that `datetime` provides other useful classes like `timedelta`, although i guess you could also do `from datetime import datetime, timedelta`

dozer/cogs/ftc.py
j5155: Should be resolved in [`fe02dfb` (#460)](https://github.com/FRCDiscord/Dozer/pull/460/commits/fe02dfb44dc8b81d3958446b75ae45c775aea5e7)

dozer/cogs/ftc.py
```diff
@@ -230,6 +271,19 @@ async def team(self, ctx: DozerContext, team_num: int):
                         value=', '.join((team_data['city'], team_data['stateProv'], team_data['country'])) or "Unknown")
             e.add_field(name='Org/Sponsors', value=team_data.get('nameFull', "").strip() or "_ _")
             e.add_field(name='Website', value=website or 'n/a')
+            e.add_field(name='Team Info Page', value=f'https://ftcscout.org/teams/{team_num}')
```
guineawheek, line 274: this seems redundant with the url link in the embed title. Maybe clarify this is FTCScout?

dozer/cogs/ftc.py
```diff
@@ -230,6 +271,19 @@ async def team(self, ctx: DozerContext, team_num: int):
...
+
+            if sres.status != 404:
```
guineawheek, line 276: tbh i would separate out OPR information to a separate command (maybe `%ftc opr`). I don't think it really belongs in the general command (and could get kinda toxic really quickly)

dozer/cogs/ftc.py
```diff
@@ -230,6 +271,19 @@ async def team(self, ctx: DozerContext, team_num: int):
                         value=', '.join((team_data['city'], team_data['stateProv'], team_data['country'])) or "Unknown")
             e.add_field(name='Org/Sponsors', value=team_data.get('nameFull', "").strip() or "_ _")
             e.add_field(name='Website', value=website or 'n/a')
```
guineawheek, line 273: the `get_none_strip()` function is broken because `team_data['website']` can be `None` and that gets turned into `str(None)`. it's not strictly required but i would recommend fixing this by rewriting it to be:
```python
def get_none_strip(s, key):
    """Ensures that a get always returns a stripped string.""" 
    return str(s.get(key, "") or "").strip()
```

guineawheek (changes requested):

dozer/cogs/ftc.py
j5155: This is a relic of the toa command, agree with you though

dozer/cogs/ftc.py
j5155: How will this become toxic?
Not sure how making it a dedicated command would help if so
This suggestion received overwhelming support in FTC discord (29 up 0 down), maybe discuss it there? You might be right I'm just not sure I understand what you mean

j5155:
Will try to address those issues tonight @guineawheek

skruglov2023:
Believe the latest comments have been resolved via latest merge with this patch branch

dozer/cogs/ftc.py
guineawheek: >  This suggestion received overwhelming support in FTC discord (29 up 0 down), maybe discuss it there? You might be right I'm just not sure I understand what you mean

First off, you said it _could_ happen, not be something that gets tacked onto every team's entry. 

> How will this become toxic?
Having OPR be something the bot pulls could quickly become a way to invalidate takes that users make. It could indirectly become a way of saying "your opinion is invalid because your team is bad"

dozer/cogs/ftc.py
skruglov2023: If/when I have time I'll probably make a new PR for the kickoff case (it's still a few months until this year's championship so no point in fixing the command *yet*

dozer/cogs/ftc.py
skruglov2023: image
Seems like the main concern is specifying where the data comes from (ftc-events/ftcscout) rather than toa
/toa is probably going to remain an alias for the /ftc team command due to how used it appears to be still. 
Maybe we can keep the default OPR for now and get some feedback from the wider community, and adjust based off observations and usage?

j5155:
OPR has been seperated out into it's own command, %ftc opr, aliased to %topr (alias broken in [`142487a` (#460)](https://github.com/FRCDiscord/Dozer/pull/460/commits/142487acba60f6de0acebbeee05f460b01326bc2) but should hopefully be fixed once stephan is free)
does this address your concerns @guineawheek ?

guineawheek:
i think opr is fine separately

skruglov2023:
%topr and %ftc opr should now be up to date on this PR, as per @j5155's message

j5155:
This PR is ready to merge as far as my planned code changes go.
@devyntk looks like GitHub is still waiting for your review. We changed to f strings as you suggested, are there any other changes you would like?

devyntk (approved):

PR #470, opened 2024-03-21
Moved attacks list to separate file

silvanathecat:
this makes it easier to add and remove them. i also added a meme one about the chain this year just because i needed to change the file for the commit to work.

ignore my last pr, the `attacks.py` file wasn't there.

### DIFF
--- a/docs/Levels.rst
+++ b/docs/Levels.rst
@@ -3,7 +3,7 @@ Levels
 ======
 meesyncs
 ++++++++
-Function to scrap ranking data from the mee6 api and save it to the
+Function to scrape ranking data from the mee6 api and save it to the
 database
 ::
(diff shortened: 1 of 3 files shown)


silvanathecat:
haven't intrinsically tested it, but it should be fine... as long as i included all the required files...

devyntk:
Im not sure how necessary this is.. I could see a much better argument for making the attacks a static variable at the top of the file, but a separate file for a ~200 line file to get smaller seems like overkill to me. If we're insistent on separating this out, it should at least be a static file (json or toml is my preference) that's not in source for easy customization.

silvanathecat:
will rewrite for json on my lunch... ive also been thinking about possibly adding a command to add attacks and a better system for damage, will probably work on that... tomorrow? busy tonight after class, but could do it late.

PR #472, opened 2024-04-01
Fix a long standing bug that I somehow haven't noticed 

tweirtx:


tweirtx:
It's a one line change I tested on my machine, committing aboose

PR #475, opened 2024-06-03
Update FRC/FTC blogs to new URLs

devyntk:


tweirtx (approved):
After work, I can see about disabling subscriptions to the blog, starting it up and letting it process, and then turning subscriptions back on once it won't be spammy.

devyntk:
Actually, nevermind, I forgot how my own code works. On restart, it'll index all the currently-visible GUIDs and won't post those, so it will only post the new posts from that point onwards. Safe to merge without coordination.

PR #476, opened 2024-06-06
shortcuts.py update

skruglov2023:
Adding fuzzy shortcut usage as well as support for shortcuts in messages

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "githubPullRequests.ignoredPullRequestBranches": [
+        "master"
+    ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "githubPullRequests.ignoredPullRequestBranches": [
-        "master"
-    ]
-}
--- a/dozer/cogs/shortcuts.py
+++ b/dozer/cogs/shortcuts.py
@@ -143,35 +143,35 @@ class Shortcuts(Cog):
         prefix = setting.prefix
         prefix_index = msg.content.find(prefix)
 
+
         if prefix_index != -1:
             # before running any chatgpt stuff, check if there's a space between prefix and shortcut
             if prefix_index + len(prefix) < len(msg.content) and msg.content[prefix_index + len(prefix)] == ' ':
                 return  # there's a space, so it was probably meant to be used in text rather than call a shortcut
-            # Extract the command part from the message directly after the prefix
+            # Extract the word immediately after the prefix
             start_index = prefix_index + len(prefix)
             remaining_content = msg.content[start_index:].strip()
+            first_word = remaining_content.split()[0] if remaining_content else ''
 
-            # Check if the remaining content starts with a valid command name
-            all_shortcuts = await ShortcutEntry.get_by(guild_id = msg.guild.id)
+            # Check if the first word is a valid command name
+            all_shortcuts = await ShortcutEntry.get_by(guild_id=msg.guild.id)
             all_shortcuts = [s.name for s in all_shortcuts]
 
-            best_match = process.extractOne(remaining_content, all_shortcuts, scorer = fuzz.partial_ratio)
+            best_match = process.extractOne(first_word, all_shortcuts, scorer=fuzz.partial_ratio)
 
-            if best_match and best_match[1] > 90:  # Adjust the threshold as needed
+            if best_match and best_match[1] > 80:  # Adjust the threshold as needed
                 shortcut_name = best_match[0]
 
-                # Ensure the command follows immediately after the prefix without any intervening characters
-                if remaining_content.startswith(shortcut_name.casefold()):
-                    shortcut = await ShortcutEntry.get_unique_by(guild_id = msg.guild.id, name = shortcut_name)
-                    if msg.reference:
-                        # Fetch the original message being replied to
-                        original_message = await msg.channel.fetch_message(msg.reference.message_id)
-                        if original_message:
-                            # Ping the original author in the new message
-                            await original_message.reply(f"{shortcut.value}")
-                    else:
-                        # Send the shortcut value without pinging if original message is not found
-                        await msg.channel.send(shortcut.value)
+                shortcut = await ShortcutEntry.get_unique_by(guild_id=msg.guild.id, name=shortcut_name)
+                if msg.reference:
+                    # Fetch the original message being replied to
+                    original_message = await msg.channel.fetch_message(msg.reference.message_id)
+                    if original_message:
+                        # Ping the original author in the new message
+                        await original_message.reply(f"{shortcut.value}")
+                else:
+                    # Send the shortcut value without pinging if original message is not found
+                    await msg.channel.send(shortcut.value)
         else:
             # If the prefix is not found in the message, do nothing
             pass
--- a/dozer/cogs/shortcuts.py
+++ b/dozer/cogs/shortcuts.py
@@ -3,7 +3,6 @@
 import discord
 from discord.ext import commands
 from discord.ext.commands import BadArgument, guild_only, has_permissions
-# from fuzzywuzzy import process, fuzz
 
 from dozer.context import DozerContext
 from ._utils import *
@@ -143,9 +142,6 @@ class Shortcuts(Cog):
         prefix = setting.prefix
         prefix_index = msg.content.find(prefix)
 
-        # Check for the presence of the prefix anywhere in the message
-        prefix_index = msg.content.find(prefix)
-
         if prefix_index != -1:
             # Ensure there's no space directly after the prefix
             if prefix_index + len(prefix) < len(msg.content) and msg.content[prefix_index + len(prefix)] == ' ':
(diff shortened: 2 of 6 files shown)


dozer/cogs/shortcuts.py
```diff
@@ -129,29 +130,51 @@ async def list(self, ctx):
...
+                return  # there's a space, so it was probably meant to be used in text rather than call a shortcut
+            # Extract the command part from the message directly after the prefix
+            start_index = prefix_index + len(prefix)
+            remaining_content = msg.content[start_index:].strip()
+
+            # Check if the remaining content starts with a valid command name
+            all_shortcuts = await ShortcutEntry.get_by(guild_id = msg.guild.id)
+            all_shortcuts = [s.name for s in all_shortcuts]
+
+            best_match = process.extractOne(remaining_content, all_shortcuts, scorer = fuzz.partial_ratio)
```
abidingabi, line 158:
```python
            best_match = process.extractOne(remaining_content, all_shortcuts, scorer = fuzz.partial_ratio)
```
This is a logic bug, if a message contains e.g.
`gm0 is neat, !roadrumner`, this will find gm0 and say there is no match, because gm0 does not contain a prefix before it.

```py
prefix_index = msg.content.find(prefix)
```
is also a logic bug, as `hi! you should check out !rr` would fail because `find` would find the first `!` (assuming `!` is the prefix.

I would recommend first finding substrings of the form `!(\w|\d|-|_`)+` or so (probably not with regex) (where ! is the server-specific prefix), but honestly this feels like a very large amount of computation to run on every message compared to checking if it starts with a specific substring, for not all that much benefit.

skruglov2023: So the first one would actually work, as it's only checking after prefix. the second one is a bug though, but then not every server uses ! as their shortcuts prefix, that's mainly a ftc thing. It's also set to not check for shortcuts if there's a space/nothing right after the prefix. If there's more checks for prefixes, however, we might as well just add something to check every instance of the prefix being present, and bring in potential for calling multiple shortcuts in a single message. That, however, opens up a window for increasing computation as it would be possible to send a message with `!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!` or similar as the text. 

That comes down to this: is it worth adding more checks for additional prefixes in the message? Or is checking if there's a shortcut after the first one enough.

.vscode/settings.json
abidingabi, line 1: this file should be removed

requirements.txt
abidingabi, line 1: there should be no changes to the requirements.txt

dozer/cogs/shortcuts.py
```diff
@@ -129,29 +130,52 @@ async def list(self, ctx):
...
-        for shortcut in shortcuts:
-            if c.lower()[len(setting.prefix):] == shortcut.name.lower():
-                await msg.channel.send(shortcut.value)
-                return
+        # Search for the prefix within the message content
+        prefix = setting.prefix
+        prefix_index = msg.content.find(prefix)
+
+        # Check for the presence of the prefix anywhere in the message
+        prefix_index = msg.content.find(prefix)
```
abidingabi, line 147: Does this not just do the same thing twice?
```suggestion
        # Check for the presence of the prefix anywhere in the message
        prefix_index = msg.content.find(prefix)
```

dozer/cogs/shortcuts.py
```diff
@@ -3,6 +3,7 @@
 import discord
 from discord.ext import commands
 from discord.ext.commands import BadArgument, guild_only, has_permissions
+# from fuzzywuzzy import process, fuzz
```
abidingabi, line 6:
```suggestion
```

dozer/cogs/shortcuts.py
```diff
@@ -129,29 +130,52 @@ async def list(self, ctx):
...
-                await msg.channel.send(shortcut.value)
-                return
+        # Search for the prefix within the message content
+        prefix = setting.prefix
+        prefix_index = msg.content.find(prefix)
+
+        # Check for the presence of the prefix anywhere in the message
+        prefix_index = msg.content.find(prefix)
+
+        if prefix_index != -1:
```
abidingabi, line 149: Since we only check one prefix index, a message like
> Hi! Check out !gm1.
would not work properly. We should iterate over all possible (non-overlapping) matches. This also would allow the use of multiple shortcuts in one message.

dozer/cogs/shortcuts.py
```diff
@@ -129,29 +130,52 @@ async def list(self, ctx):
...
+            # Extract the command part from the message directly after the prefix
+            start_index = prefix_index + len(prefix)
+            remaining_content = msg.content[start_index:].strip()
+
+            # Extract the command name before any spaces or punctuation
+            command_name = remaining_content.split()[0] if remaining_content.split() else ""
+
+            # Check if the extracted command name matches any valid shortcuts
+            all_shortcuts = await ShortcutEntry.get_by(guild_id = msg.guild.id)
+            all_shortcuts = [s.name for s in all_shortcuts]
```
abidingabi, line 163: Why do this? We need both the name and value information in the loop.

dozer/cogs/shortcuts.py
```diff
@@ -129,29 +130,52 @@ async def list(self, ctx):
...
+
+            if command_name in all_shortcuts:
+                shortcut = await ShortcutEntry.get_unique_by(guild_id = msg.guild.id, name = command_name)
```
abidingabi, line 166: If we change to use the full shortcut entity in the loop, we do not need to do more database calls. even if we needed to, this should only be done in the case where we actually need the shortcut value.

dozer/cogs/shortcuts.py
```diff
@@ -129,29 +130,52 @@ async def list(self, ctx):
...
+
+            # Extract the command part from the message directly after the prefix
+            start_index = prefix_index + len(prefix)
+            remaining_content = msg.content[start_index:].strip()
+
+            # Extract the command name before any spaces or punctuation
+            command_name = remaining_content.split()[0] if remaining_content.split() else ""
+
+            # Check if the extracted command name matches any valid shortcuts
+            all_shortcuts = await ShortcutEntry.get_by(guild_id = msg.guild.id)
```
abidingabi, line 162: The names should be sorted by reverse length, otherwise we have issues when we have two shortcuts that are prefixes of one another; consider `!foo` and `!foobar`, with `!foo` showing up first in the list. That can be achieved by writing an SQL query with an `ORDER BY` clause.

dozer/cogs/shortcuts.py
```diff
@@ -129,29 +130,52 @@ async def list(self, ctx):
...
+            if command_name in all_shortcuts:
+                shortcut = await ShortcutEntry.get_unique_by(guild_id = msg.guild.id, name = command_name)
+                if msg.reference:
+                    # Fetch the original message being replied to
+                    original_message = await msg.channel.fetch_message(msg.reference.message_id)
+                    if original_message:
+                        # Ping the original author in the new message
+                        await original_message.reply(f"{shortcut.value}")
+                else:
+                    # Send the shortcut value without pinging if original message is not found
```
abidingabi, line 174:
```suggestion
                    # Send the shortcut value as a non-reply message if msg is not a reply 
```

dozer/cogs/shortcuts.py
```diff
@@ -129,29 +130,52 @@ async def list(self, ctx):
...
+                    await msg.channel.send(shortcut.value)
+        else:
+            # If the prefix is not found in the message, do nothing
+            pass
```
abidingabi, line 178:
```suggestion
```
This does nothing.

abidingabi (changes requested):

dozer/cogs/shortcuts.py
```diff
@@ -129,29 +130,52 @@ async def list(self, ctx):
...
+            # Ensure there's no space directly after the prefix
+            if prefix_index + len(prefix) < len(msg.content) and msg.content[prefix_index + len(prefix)] == ' ':
+                return  # there's a space, so it was probably meant to be used in text rather than call a shortcut
+
+            # Extract the command part from the message directly after the prefix
+            start_index = prefix_index + len(prefix)
+            remaining_content = msg.content[start_index:].strip()
+
+            # Extract the command name before any spaces or punctuation
+            command_name = remaining_content.split()[0] if remaining_content.split() else ""
```
abidingabi, line 159:
```suggestion
            message_segments = remaining_content.split()
            if not message_segments:
                return
            command_name = message_segments[0]
```

dozer/cogs/shortcuts.py
abidingabi, line 166: `get_unique_by` does not exist.

abidingabi, line 166: This can all be replaced with something to the effect of
```suggestion
            shortcuts = await ShortcutEntry.get_by(guild_id = msg.guild.id, name = command_name)
            if not shortcuts:
                return
            shortcut = shortcuts[0]
```

dozer/cogs/shortcuts.py
skruglov2023: Are you sure? What if someone has like 5 ! or something in their message? Or, if they just want to spam like 50 ! ?

abidingabi: :woman_shrugging: it’d be a weird edge case to not work. If someone put 50 exclamation marks, it will lag things a bit but probably not be too bad? Maybe add a cap at like 10 prefixes checked or something.

dozer/cogs/shortcuts.py
skruglov2023: Must've forgotten to PR that section as well. It does exist in dozer2 (and adding it here as well)

skruglov2023:
I'll just restart the PR at this point, looks like a few random things got mixed in.

PR #490, opened 2025-11-30
Update FTC Team List for Name Game

0verkil:
Had some fun experimenting with Name Game in the ftc discord recently. Unfortunately we quickly noticed the teams list was horrifically out of date. It would be great to make it work through api instead of whatever this is but this should serve as a temporary update since I did not have time to update to API today. 

Data generated from FTCScout's database of FTC teams and processed through python. Feel free to check the file for yourself.


devyntk (dismissed):
Seems fine to me. Your pickled data seems to include several of the pre-registration teams (like 201902089) that would likely be better not to include, but I'm not going to hold up this PR for it.

0verkil:
i can modify it if needed but it really just did not seem very important

PR #491, opened 2025-12-02
Fix slash command timeouts with modlog, update documentation, and fix lint

tweirtx:
Sorry for sending such a massive PR all at once, but it's about time the dust gets blown off of this code.


WaterGame2023 (approved):

WaterGame2023:
Looks good as far as I can tell. Everything seems to check out fine

